Issue #81, opened 2018-03-15
Simplify EleventyConfig.addPassthroughCopy (via ability to chain method calls or a plural version).


kleinfreund:
It could be useful to have [`EleventyConfig`](
https://github.com/11ty/eleventy/blob/master/src/EleventyConfig.js)’s methods return a reference to itself so that method calls can be chained. This is best demonstrated with an example:

Current implementation:

```js
module.exports = function (eleventyConfig) {
  eleventyConfig.addPassthroughCopy('img');
  eleventyConfig.addPassthroughCopy('css');
  eleventyConfig.addPassthroughCopy('js');
};
```

Possible future implementation:

```js
module.exports = function (eleventyConfig) {
  eleventyConfig
    .addPassthroughCopy('img')
    .addPassthroughCopy('css')
    .addPassthroughCopy('js');
};
```

That said, it would still be necessary to repeat `addPassthroughCopy` for each path.

This could be simplified by adding a *plural* version `addPassthroughCopies` with one rest argument `paths`. It would just iterate over `paths` and call `addPassthroughCopy` for each entry. Then, for the example above, chaining would not be very helpful.

That’s just an idea :tm:. I can submit a pull request that implements either of the suggestions.

zachleat:
I like this:

```
module.exports = function (eleventyConfig) {
  eleventyConfig
    .addPassthroughCopy('img')
    .addPassthroughCopy('css')
    .addPassthroughCopy('js');
};
```

I’d be happy to review a PR that added that functionality!

zachleat:
Fixed by https://github.com/11ty/eleventy/pull/82

nhoizey:
@zachleat I see this has been merged with the chainable syntax.

Could you explain why you prefered this over a "simpler" array of resources to copy?

This feels even better:

```javascript
module.exports = function (eleventyConfig) {
  eleventyConfig.addPassthroughCopy(['img', 'css', 'js']);
};
```

PR #82, opened 2018-03-16
Make EleventyConfig.addPassthroughCopy chainable

kleinfreund:
_(Implements #81)_

In this pull request:

- `EleventyConfig.addPassthroughCopy` returns a reference to `this` (the `EleventyConfig` object in this context)
- Tests for `EleventyConfig.addPassthroughCopy` are added

Should this be mentioned in [Pass-through File Copy: Manual Passthrough Copy (Faster)](https://github.com/11ty/eleventy/blob/master/docs/copy.md#manual-passthrough-copy-faster)? I could add a little example in the code snippet.

Also, you might not like the method documentation I added. However, I think it’s very helpful as some editors actually display this in a useful manner on hovering over function calls (e.g. VS Code).

### DIFF
--- a/src/EleventyConfig.js
+++ b/src/EleventyConfig.js
@@ -154,8 +154,19 @@ class EleventyConfig {
     pluginCallback(this);
   }
 
+  /**
+   * Adds a path to a file or directory to the list of pass-through copies
+   * which are copied as-is to the output.
+   *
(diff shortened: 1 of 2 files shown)


zachleat:
Amazing—thank you!

and I do like the method documentation!

PR #96, opened 2018-04-04
Fixes #95 - Prevent partial matches when filtering tag(s).

### LINKED ISSUE #95: `collections.nav` contains partial matches for “nav”
Items I have flagged with the tag `subnav` are showing up in `collections.nav`. It looks like anything containing `nav` gets piled in there:

**about/index.md**
```markdown
layout: layouts/post.njk
title: About Me
tags: knave
```

Then shows up in:

**layouts/base.njk**
```html
{%- for nav in collections.nav | reverse -%}
	{{ nav.url | url }}
{%- endfor -%}
```

While this has clear ramifications for blogs dedicated to the British Royal Navy, I’m guessing the `subnav` thing is also gonna come up from time to time. I’m guessing this needs to be tuned up to split on spaces, then exact-match each separate string: https://github.com/11ty/eleventy/blob/master/src/TemplateCollection.js#L52

Inbox willing and the creek don’t rise, I can probably swing that at some stage.

Wilto:


zachleat:
oh dang, nice catch—thanks!

PR #101, opened 2018-04-10
Hotfix ejsoptions

### LINKED ISSUE #100: eleventyConfig.setEjsOptions is not a function
It looks like `setEjsOptions` was added to `src/engines/ejs.js` but not to `src/eleventyConfig.js` like `setPugOptions` was.

chrisdmacrae:
Fixes #100 

This adds the EJS option functions and data to EleventyConfig so it can be used.

It also normalize how this is done across Pug and Liquid as well.


zachleat:
Thank you!

PR #102, opened 2018-04-11
Feature ejs relative includes

### LINKED ISSUE #103: Add support for relative EJS includes
EJS has two types of includes:

- *Relative includes*: includes resolved from the current template being rendered.
- *Root includes*: includes resolved from the root of the project, or from a specified includes root _(in eleventy, this is `_includes/`)_

Currently Eleventy supports the latter, but not the former, with all relative includes being resolved from the root `_includes/` folder.

For example:

```
|- views/
|--- _includes/
|----- component.example
|--- component.example
|--- component.md
```

`component.md`:
```
---
title: Component
---
<!-- This is a relative require; should resolve `views/component.example` -->
<%- include("component.example"); -%>

<!-- This is a root require; should resolve `views/_includes/component.example` -->
<%- include("/component.example"); -%>
```

I've submitted a rough PR that outlines how to update the markdown and EJS engines to properly support both types, w/o hurting the Eleventy `_includes/` functionality.

This also doesn't effect the `layout` front matter field, so I think it works fairly well.

#102

chrisdmacrae:
Closes #103

Adds support for relative includes in EJS templates.


chrisdmacrae:
Need to review why the markdown tests are failing -- they work for me locally

src/TemplateRender.js
```diff
@@ -149,12 +149,19 @@ TemplateRender.prototype.render = async function(str, data) {
   return this.engine.render(str, data);
 };
 
-TemplateRender.prototype.getCompiledTemplate = async function(str) {
+TemplateRender.prototype.getCompiledTemplate = async function(str, inputPath) {
   // TODO refactor better, move into TemplateEngine logic
   if (this.engineName === "md") {
-    return this.engine.compile(str, this.parseMarkdownWith, !this.useMarkdown);
+    return this.engine.compile(
```
kleinfreund: You’re calling `compile` assuming an incorrect signature. The signature is:

```js
compile(str, inputPath, preTemplateEngine, bypassMarkdown)
```

But you pass `inputPath` last.

chrisdmacrae: Yup, that was it!

zachleat:
Hey @chrisdmacrae, do you think you could add a test for this to https://github.com/11ty/eleventy/blob/master/test/TemplateRenderEJSTest.js#L17 ? Should have some root include examples in there you can build from

zachleat:
Actually looks like there is some commented out tests here that might be worth salvaging https://github.com/11ty/eleventy/blob/master/test/TemplateRenderEJSTest.js#L53

zachleat:
Just an FYI if this makes it in today it’ll go out with the small maintenance release I’m planning to do later tonight!

chrisdmacrae:
@zachleat I'll see what I can do!

chrisdmacrae:
@zachlet tests for relative includes added. I duplicated the existing tests and ensured that it was running relative includes from the template's context.

Working like a charm.

Cheers 🍻

zachleat:
Thank you!

PR #112, opened 2018-04-23
Use ES6 syntax and async await in lieu of Promise

jabranr:
> First of all apologise for a late come back and send this PR across without discussing it beforehand. This is open for debate/suggestions/declines so feel free to point to anything that is not suitable. 💯 

Finally got some time to do some more pending work related to my previous PR. Main updates are:

* Use ES6 syntax
* Use `async await` in lieu of `Promise`
* Use `ES6 string concatenation` instead of legacy `+` based
* Added `nodejs` prerequisite to `package.json` since that is specified in docs anyway

If it goes ahead, I will then continue to work on more. 👍

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -1,64 +1,77 @@
 #!/usr/bin/env node
+
+const chalk = require("chalk");
+const argv = require("minimist")(process.argv.slice(2));
+const EleventyNodeVersionCheck = require("./src/VersionCheck");
+const Eleventy = require("./src/Eleventy");
+const EleventyCommandCheck = require("./src/EleventyCommandCheck");
+
 if (process.env.DEBUG) {
   require("time-require");
 }
 
-const argv = require("minimist")(process.argv.slice(2));
-const EleventyNodeVersionCheck = require("./src/VersionCheck");
-
-EleventyNodeVersionCheck().then(
-  function() {
-    const Eleventy = require("./src/Eleventy");
-    const EleventyCommandCheck = require("./src/EleventyCommandCheck");
-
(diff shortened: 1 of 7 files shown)


zachleat:
I kept cmd.js Node 6 compatible so that I could properly error when users try to run eleventy with node 6. I see you added node 8 to the package.json. Forgive my ignorance here—does this literally install node 8 when it isn’t available?

jabranr:
@chrisdmacrae True, sadly. Though Eleventy does not use bluebird or any third-party library for `Promise` as far as I am aware and relies on native `Promise`. Is there a plan to switch as such? 

I think I can make the use of `async await` more performant with few more changes. I will push an update later and would love your feedback on it. 

Also, are you aware if we already have any performance tests on Eleventy?

zachleat:
@jabranr We’re on the cowpaths here with native async/await and we aren’t transpiling anything so I wouldn’t worry about it. We’ll leave that level of optimization to the node core team.

I do like the idea of having a standard suite of perf tests though. I’ll make an issue.

zachleat:
Hmm, maybe I wasn’t clear enough here. I want to be able to:

```
nvm use 6
npm install -g @11ty/eleventy
eleventy
```

Current behavior in the master branch is that it will show the following error when you run eleventy:

```
Eleventy requires Node version 8 or above. You’re currently using 6.12.3
```

I would like to maintain this behavior. As such that means that we need to remove all async/await behavior from inside of `cmd.js` to get this merged and just use Promises (for only this single cmd.js file).

zachleat:
Please reopen if you circle back here!

PR #120, opened 2018-05-11
Use Browsersync for --serve

### LINKED ISSUE #119: Switch to Browsersync for --serve
Using browsersync will give us automatic reloads when the build finishes (and really nice CSS injection without reload).

zachleat:
Switch to Browsersync for --serve (seamless reloads when watch finishes). Works transparently with `pathPrefix` to host your site in the subdirectory specified in your config file.

Smart reloads CSS files:

1. If they are in _includes, reload the entire page (css files included directly in <style>).
2. If they are anywhere else, injects them without reload.

Fixes #119

### DIFF
--- a/cmd.js
+++ b/cmd.js
@@ -36,16 +36,7 @@ EleventyNodeVersionCheck().then(
         console.log(elev.getHelp());
       } else if (argv.serve) {
         elev.watch().then(function() {
-          const serve = require("serve");
-          const server = serve(elev.getOutputDir(), {
-            port: argv.port || 8080,
-            ignore: ["node_modules"]
-          });
-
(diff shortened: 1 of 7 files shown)


zachleat:
Yeah I'm trying to be clear and use browsersyncs lingo with reload and inject. Rather, any change to files that are watched will trigger a reload (full browser reload) and changes to CSS files outside of _includes will inject the new stylesheets (not a reload, but changes will apply automatically). CSS files will need to be in your template format list or in passthrough copy files/directories.

zachleat:
For the record I'd still like to add configuration API methods to extend this. But the rest of what you're asking about should be handled transparently (except the https stuff)

PR #122, opened 2018-05-12
Update liquidjs dependancy

paulrobertlloyd:
Hey @zachleat! Not sure if there was a particular reason for keeping liquidjs pinned to this earlier version; all the tests pass updating to v4.0.0.

Recent updates to this dependancy fix a number of issues, not least [forloop.last being incorrect if the loop had a limit applied](https://github.com/harttle/liquidjs/issues/67), and the [`strip_html` filter not working](https://github.com/harttle/liquidjs/issues/68) in a way consistent with Ruby Liquid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "gray-matter": "^3.1.1",
     "hamljs": "^0.6.2",
     "handlebars": "^4.0.11",
-    "liquidjs": "^2.2.1",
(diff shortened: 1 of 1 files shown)


zachleat:
Nice! I'll have a look. For the record there are some skipped tests in there that I was hoping would work after the update. But if nothing regresses, no harm in upgrading 👍🏻

PR #201, opened 2018-08-20
Update liquid to ^5.1.0

jdsteinbach:
I recently found a bug in liquidjs (that required me to rewrite a template file that had worked with the liquid gem when I used Jekyll) and it was patched here: https://github.com/harttle/liquidjs/issues/72

I manually added the better liquidjs version to my local site & haven't found any regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "gray-matter": "^4.0.1",
     "hamljs": "^0.6.2",
     "handlebars": "^4.0.11",
-    "liquidjs": "^4.0.0",
+    "liquidjs": "^5.1.0",
     "lodash.chunk": "^4.2.0",
     "lodash.clone": "^4.5.0",
     "lodash.get": "^4.4.2",


zachleat:
Thanks!

zachleat:
Whoa I would not have expected `{% include {{test}} %}` to work in any case—huh.

kleinfreund:
> Whoa I would not have expected {% include {{test}} %} to work in any case—huh.

Me neither, but then again, how would you use a variable without `dynamicPartials`?

zachleat:
@kleinfreund I was of the opinion that you, uh, couldn’t?

jdsteinbach:
Before 11ty (Liquid Ruby gem via Jekyll), I was using this code:

```
{% assign rand = page.title | size | modulo: 5 | plus: 1 %}
{% include bg-{{rand}}.html %}
```

That broke in liquidjs (prior to the update in this PR) so I refactored to a `switch` (too verbose imo, but it worked): https://github.com/jdsteinbach/jdsteinbach.github.io/blob/blog/_includes/header-image.liquid

I wasn't aware of that `dynamicPartials` option though - very cool! I'll look for some time soon to test things out & see if that option was what I actually needed. Thanks!

kleinfreund:
@zachleat Well, using `dynamicPartials` is what I need then. :)

@jdsteinbach Maybe something like this works. Not very beautiful, though.

```liquid
{% assign rand = page.title | size | modulo: 5 | plus: 1 %}
{% assign path =  rand | prepend: 'bg-' | append: '.html' %}
{% include path %}
```

zachleat:
Relevant docs: https://www.11ty.io/docs/languages/liquid/

Issue #173, opened 2018-07-14
Support `--input=glob` on the command line


zachleat:
Right now we support:

```
--input=.
--input=otherdirectorypath/maybeasubdir
--input=filename.md
--input=maybeasubdir/filename.md
```

But maybe add support for

```
--input=*.md
--input=maybeasubdir/*.md
```

zachleat:
Thank you @harmenjanssen!

PR #276, opened 2018-10-15
Check whether file exists before isDirectory check

harmenjanssen:
This adds a file exists check before statting the file.
This avoids an exception being thrown, but more importantly, allows glob
arguments to be passed as `--input`, fixing #173 in the process.

An example of a glob pattern that didn't work before, but will after this change, is:

```
$ eleventy --input=pages/!(contact.md)
```

It will process all file in the `pages` directory, except `contact.md`.

I've added a unit test for the glob argument, let me know if you'd like to see more checks!


zachleat:
Thank you!!

PR #363, opened 2019-01-08
Fix: Tests invoking illegal constructor TemplateData() (fixes #362)

### LINKED ISSUE #362: Tests invoke illegal constructor TemplateData()
In [`test/TemplateTest.js`](https://github.com/11ty/eleventy/blob/master/test/TemplateTest.js) and [`test/TemplateDataTest.js`](https://github.com/11ty/eleventy/blob/master/test/TemplateDataTest.js), the following is used:

```js
let dataObj = new TemplateData();
```

However, `TemplateData()` takes one argument (`inputData`). This leads to an unexpected data directory path starting with `undefined/`. This is not caught by any of the tests and creates problems for #231.

Possible solutions:

- Never invoke `TemplateData()` in the tests without its required arguments
- Assign a default value to its argument `inputDir`

Since in regular Eleventy application, `TemplateData()` is never invoked without an argument, the tests shouldn’t either. Adding a path like this …

```js
let dataObj = new TemplateData("./");
```

… doesn’t break any tests and avoids the illegal invokation of the constructor, too.

(This is a development-specific bug with Eleventy’s tests.)

kleinfreund:
This pull requests:

- replaces all instances of `TemplateData()` in the tests with one of the following:
  - `TemplateData("./")` when targetting the global (or project) directory
  - `TemplateData("./test/")` when targetting a directory that doesn’t have a `_data` directory inside
  - `TemplateData("./test/stubs/")` in all other cases

This fixes #362.

### DIFF
--- a/test/TemplateDataTest.js
+++ b/test/TemplateDataTest.js
@@ -102,7 +102,7 @@ test("addLocalData() doesn’t exist but doesn’t fail (template file does not 
 });
 
 test("Global Dir Directory", async t => {
-  let dataObj = new TemplateData();
+  let dataObj = new TemplateData("./");
 
   t.deepEqual(await dataObj.getGlobalDataGlob(), ["./_data/**/*.(json|js)"]);
 });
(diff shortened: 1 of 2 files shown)


zachleat:
Thank you!

Issue #368, opened 2019-01-09
RFC: Restrict Eleventy’s access to directories outside of its realm


kleinfreund:
Currently, Eleventy does not restrict access to directories outside of its realm. This means that both input and output directory can be outside of Eleventy’s project directory (i.e. the place from where you run `eleventy`, a.k.a. the command directory). The following is possible:

```sh
eleventy --input=../private --output=../other-project
```

This is dangerous and a potential security risk for the following reasons:

- Users might override their own data by accident.
- In a hosted environment (see [jekyllrb.com: Security Fixes for series 3.6, 3.7 and 3.8](https://jekyllrb.com/news/2018/09/19/security-fixes-for-3-6-3-7-3-8/)), this would be an immediate non-starter.

I propose disallowing Eleventy from reading/writing outside of its root/project/command directory.

zachleat:
yeah, I’m on board with this. `--input` is an easier sell than `--output` I think. But I’m on board with both.

Would this also require a review of any templating `include` functionality?

kleinfreund:
Algorithm:

1. Normalize and strip leading double-dot segments from path
2. Use [Node.js’ fs.realpath](https://nodejs.org/api/fs.html#fs_fs_realpath_path_options_callback) to check …

   - if the path doesn’t exist, throw an error.
   - if path is outside of Eleventy’s realm, throw an error.

   Otherwise, return the absolute path.

This algorithm allows usage of symbolic links as long as they point inside Eleventy’s realm while restricting access to any directory outside of it.

Node.js implementation (draft) for step 2:

```node
const fs = require('fs');

function safePath(path) {
  try {
    const realPath = fs.realpathSync(path);

    if (realPath.startsWith(process.cwd())) {
      return path;
    }

    throw new Error(`Path ${realPath} is outside of Eleventy’s project directory.`);
  } catch (error) {
    throw new Error(error.path ? `Path ${error.path} does not exist.` : error);
  }
}
```

zachleat:
@kleinfreund `_includes` should be completely optional unless you try to use it for something.

aral:
Unless I’m mistaken, this change will make it much harder (if not impossible) to include eleventy in any other command-line apps (e.g., https://sitejs.org) that take a path to operate on:

```sh
site path/to/site/to/serve
```

If, say, eleventy was included in Site.js, we’d have to make sure that the only way the server was ever run was:

```sh
cd path/to/site/to/serve
site
```

(Unless I am missing something.)

[...]

PR #376, opened 2019-01-11
Document & refactor the TemplatePath utilities

kleinfreund:
Before tackling #368, I realized I needed to know more about `TemplatePath`’s utility functions. I went ahead and added documentation comments to every function. I also refactored some of them, mainly with the goal of simplifying the code.

I created a separate commit for each function because this makes it way easier to review the whole thing (e.g. function name changes are happening in the commit where that function is concerned). Should this get merged, you potentially want to squash all commits with the prefix `TemplatePath: ` into one. Your choice.

## Changes

- `TemplatePath._getModuleDir`: <del>prefixed the function name with an underscore as it’s done in other files to signal that it’s only used in tests.</del><ins>See https://github.com/11ty/eleventy/pull/376#issuecomment-453755611</ins>
- `TemplatePath.getWorkingDir`: <del>Normalizing the working directory so all code is dealing with the same.</del><ins>See https://github.com/11ty/eleventy/pull/376#issuecomment-453755611</ins>
- `TemplatePath.getDir`: Added some tests with glob patterns.
- `TemplatePath.getDirFromFilePath`: Added tests.
- `TemplatePath.getLastPathSegment`: Renamed from `getLastDir`. Refactored for readability.
- `TemplatePath.getAllDirs`: Refactored for readability. Adjusted tests in order, not values. I wonder if there is an orderless `deepEqual`. Reading the diffs for those is extremely cumbersome.
- `TemplatePath.hasTrailingSlash`: <del>Refactored slightly to handle `undefined` input explicitly. I don’t think this should handle `undefined` at all, to be honest. Passing `undefined` here should be an immediate error.</del><ins>See https://github.com/11ty/eleventy/pull/376#issuecomment-453824902</ins>
- `TemplatePath.normalizeUrlPath`: <del>Just the documentation string. No implementation change.</del><ins>See https://github.com/11ty/eleventy/pull/376#issuecomment-453824902</ins>
- `TemplatePath.absolutePath` + `TemplatePath.relativePath`: Renamed from `localPath` and `delocalPath` (:cool:). Replaced calls to `absolutePath` without an argument with `getWorkingDir` because it’s the same and more clear.
- `TemplatePath.addLeadingDotSlash` + `TemplatePath.addLeadingDotSlashArray`:
- `TemplatePath.stripLeadingDotSlash`: Slight readability changes.
- `TemplatePath.startsWithSubPath`: Renamed from `contains` (*contains* is not the right meaning because it only matches at the start of the path)
- `TemplatePath.stripLeadingSubPath`: Renamed from `stripPathFromDir` because it wasn’t clear which argument is which when reading a call to the function. Both arguments are paths.
- `TemplatePath.isDirectorySync`: Documentation comment.
- `TemplatePath.convertToRecursiveGlob`: Renamed from `convertToGlob` to make clear what kind of glob pattern is meant. Also added missing arguments to all function calls in the tests to avoid handling `undefined` as an argument.
- `TemplatePath.getExtension`: Use Node.js’ `path.extname`.
- `TemplatePath.removeExtension`: Re-use `TemplatePath.getExtension`.

---

Okay, I admit, this is quite the massive pull request. I won’t be mad if it doesn’t get merged (or reviewed anytime soon for that matter). :sweat_smile:

### DIFF
--- a/test/TemplatePathTest.js
+++ b/test/TemplatePathTest.js
@@ -8,6 +8,15 @@ test("getWorkingDir", t => {
   t.is(TemplatePath._getModuleDir(), path.resolve(__dirname, ".."));
 });
 
+test("getDir", t => {
+  t.is(TemplatePath.getDir("README.md"), ".");
+  t.is(TemplatePath.getDir("test/stubs/config.js"), "test/stubs");
+  t.is(TemplatePath.getDir("./test/stubs/config.js"), "./test/stubs");
+  t.is(TemplatePath.getDir("test/stubs/*.md"), "test/stubs");
+  t.is(TemplatePath.getDir("test/stubs/**"), "test/stubs");
+  t.is(TemplatePath.getDir("test/stubs/!(multiple.md)"), "test/stubs");
+});
+
 test("normalize", async t => {
   t.is(TemplatePath.normalize(""), ".");
   t.is(TemplatePath.normalize("."), ".");
@@ -132,12 +141,6 @@ test("stripPathFromDir", t => {
 
   t.is(TemplatePath.stripPathFromDir(".htaccess", "./"), ".htaccess");
   t.is(TemplatePath.stripPathFromDir(".htaccess", "."), ".htaccess");
-});
-
-test("getDir", t => {
-  t.is(TemplatePath.getDir("README.md"), ".");
-  t.is(TemplatePath.getDir("test/stubs/config.js"), "test/stubs");
-  t.is(TemplatePath.getDir("./test/stubs/config.js"), "./test/stubs");
 });
 
 test("getLastDir", t => {
--- a/test/TemplatePathTest.js
+++ b/test/TemplatePathTest.js
@@ -5,7 +5,7 @@ import TemplatePath from "../src/TemplatePath";
 
 test("Working dir", t => {
   t.is(TemplatePath.getWorkingDir(), path.resolve("./"));
-  t.is(TemplatePath.getModuleDir(), path.resolve(__dirname, ".."));
+  t.is(TemplatePath._getModuleDir(), path.resolve(__dirname, ".."));
 });
 
 test("normalize", async t => {
--- a/test/TemplatePathTest.js
+++ b/test/TemplatePathTest.js
@@ -96,6 +96,22 @@ test("hasTrailingSlash", t => {
   t.is(TemplatePath.hasTrailingSlash("./test/stubs/"), true);
 });
 
+test("absolutePath", t => {
+  t.is(
+    TemplatePath.absolutePath(".eleventy.js")
+      .split("/")
+      .pop(),
+    ".eleventy.js"
+  );
+});
+
+test("absolutePath and relativePath", t => {
+  t.is(
+    TemplatePath.relativePath(TemplatePath.absolutePath(".eleventy.js")),
+    ".eleventy.js"
+  );
+});
+
 test("stripLeadingDotSlash", t => {
   t.is(TemplatePath.stripLeadingDotSlash("./test/stubs"), "test/stubs");
   t.is(TemplatePath.stripLeadingDotSlash("./dist"), "dist");
@@ -227,20 +243,4 @@ test("stripLeadingDots", t => {
   t.is(TemplatePath.stripLeadingDots("./dist"), "/dist");
   t.is(TemplatePath.stripLeadingDots("../dist"), "/dist");
   t.is(TemplatePath.stripLeadingDots("dist"), "dist");
-});
-
-test("localPath", t => {
-  t.is(
-    TemplatePath.localPath(".eleventy.js")
-      .split("/")
-      .pop(),
-    ".eleventy.js"
-  );
-});
-
-test("localPath and delocalPath", t => {
-  t.is(
-    TemplatePath.delocalPath(TemplatePath.localPath(".eleventy.js")),
-    ".eleventy.js"
-  );
 });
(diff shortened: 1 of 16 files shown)


kleinfreund:
I removed the tests for `TemplatePath.getWorkingDir()` and `TemplatePath._getModuleDir()` because they’re redundant. Also, since I didn’t normalize the expected result of the former, the test was breaking on Windows.

I also rebased against the master branch due to the `Eleventy.js` refactor and resolved any conflicts.

kleinfreund:
I refactored `TemplatePath.normalizeUrlPath` and by doing so, removed `TemplatePath.hasTrailingSlash`.

Rational: `TemplatePath.hasTrailingSlash` was only used by methods dealing with URL paths while at the same time containing code dealing with Windows path separators (i.e. `\\`). It took me a long while to realize this and figure out why this was special-cased. It turned out to be unnecessary.

`TemplatePath.normalizeUrlPath` now only replaces one or multiple trailing slashes with a single one after already being normalized by calling `path.join`. I also added some specific tests to `TemplatePathTest.js`.

As a neat side-effect, the file `TemplatePath` now has 100% coverage. Only `TemplatePath.getWorkingDir()` isn’t explicitly tested as I don’t find a way to provide a meaningful test.

I apologize for the intensive force-pushing I’m doing here. As I’m the only one comitting to my fork’s branch and there are no review notes, yet, it should be fine.

zachleat:
Just FYI I haven’t lost this! It looks amazing and I will try to get to it soon 🎉

kleinfreund:
Ah, don’t worry. I’m in the midst of applying around the globe and I’m just happy to spend some weekends with open source projects. There is no hurry on my end. :)

zachleat:
Started to look at this—I’m really excited about these changes, looks really great!

I did confirm locally that all the tests pass. However when I built the 11ty.io website using this version of the code the quick tips section didn’t render correctly.

I’m still diagnosing why but it’s missing a title and the collection list at the bottom.

image

versus building on `master`:
image

So I guess we have two action items here:
- [ ] Add a test that captures this behavior
- [ ] Fix the issue and the newly broken test

zachleat:
Looks like the order was important on `TemplatePath.getAllDirs` and it may have changed here.

zachleat:
Specifically it looks like your version has the following test results in reverse order? https://github.com/11ty/eleventy/blob/master/test/TemplatePathTest.js#L151

test/TemplatePathTest.js
```diff
@@ -1,11 +1,44 @@
...
+  t.deepEqual(TemplatePath.getAllDirs("./"), ["."]);
+  t.deepEqual(TemplatePath.getAllDirs("./testing"), ["./testing"]);
+  t.deepEqual(TemplatePath.getAllDirs("./testing/"), ["./testing"]);
+  t.deepEqual(TemplatePath.getAllDirs("testing/"), ["testing"]);
+  t.deepEqual(TemplatePath.getAllDirs("testing"), ["testing"]);
+  t.deepEqual(TemplatePath.getAllDirs("./testing/hello"), [
+    "./testing",
+    "./testing/hello"
+  ]);
+  t.deepEqual(TemplatePath.getAllDirs("./src/collections/posts"), [
```
zachleat: This expected result should be in reverse order, I think.

test/TemplateTest.js
```diff
@@ -447,12 +447,12 @@ test("Local template data file import (two subdirectories deep)", async t => {
   );
 
   t.deepEqual(await dataObj.getLocalDataPaths(tmpl.getInputPath()), [
```
zachleat: The original order for this test should be restored

kleinfreund:
Ah, I see, this has come back to bite me. I noticed that the new implementation for `TemplatePath.getAllDirs` reversed the order, but thought the order wasn’t important. Upon changing the broken tests to reflect the reversing, I was actually under the impression that `deepEqual` was not caring about the order. It does care about the order of elements in an array, but not about the order of properties on an object. I read this Ava issue wrong: https://github.com/avajs/ava/issues/1232.

---

Now do you want me to squash these amends into the respective commits? I also tried rebasing against master yesterday and it went through without conflicts. Want me to rebase and force-push?

zachleat:
Naw squashing/rebasing is unnecessary. My preference is to see the raw history 🤷‍♂️

🎉🎉🎉🎉🎉🎉

I think we’re good here—thank you so much for this refactor! It really cleaned things up nicely!

We’ll plan on this going in to 0.7.2 (or 0.8.0, whichever comes next)

zachleat:
Oh one more thing, are there any issues this resolves that need to be closed? A quick peek didn’t seem to reveal anything there, just want to make sure the issue tracker is updated if need-be

kleinfreund:
@zachleat Not that I’m aware of. This was a very much under-the-hood endeavour. I usually make notes when I notice such connections, but I don’t have any. If I notice something, I’ll comment in the appropriate places.

When I find time, I want to look into making sure that we always deal with paths that are anchored to the Eleventy root directory. That’s tracked in #368.

aral:
@kleinfreund @zachleat This merge request is currently blocking me from integrating Eleventy into [Site.js](https://sitejs.org) as Site.js accepts a path to operate on.

Is there any chance you’d consider a pull request that makes this feature optional so that Eleventy can act on directories other than its current working directory?

kleinfreund:
@aral I think you might be confusing this MR with another one: https://github.com/11ty/eleventy/issues/368

The work in here should prevent you in any way from what you describe in https://github.com/11ty/eleventy/issues/931.

Issue #353, opened 2018-12-21
Add Windows to travis-ci builds


zachleat:
https://blog.travis-ci.com/2018-10-11-windows-early-release

zachleat:
Windows samples here https://travis-ci.org/11ty/eleventy/builds/478638999

kleinfreund:
Maybe we could do something like sorting both expected and actual results.

zachleat:
Tried this again, looks a bit more straightforward this time: https://travis-ci.org/11ty/eleventy/builds/516470666

kleinfreund:
The errors are related to the `url()` function which has some code producing Windows-style path delimiters. This should never happen when dealing with URLs. Should be easy enough to fix. Unfortunately, I no longer have a Windows system available to me to test this. Maybe a fix is possible without by looking at the Travis results.

zachleat:
Looks like this was fixed in #486! Thank you @AndrewAsquith!!

zachleat:
Heck yeah https://travis-ci.org/11ty/eleventy/builds/518844789

kleinfreund:
Heck yeah indeed!

PR #381, opened 2019-01-12
Fix broken tests on Windows (yes, it’s newlines)

kleinfreund:
I booted into Windows after months again and ran Eleventy’s tests. Some were broken due to newline mismatch. I fixed them by normalizing newlines to `\n` in the tests.

### DIFF
--- a/test/TemplateMapTest.js
+++ b/test/TemplateMapTest.js
@@ -2,6 +2,7 @@ import test from "ava";
 import Template from "../src/Template";
 import TemplateMap from "../src/TemplateMap";
 import TemplateCollection from "../src/TemplateCollection";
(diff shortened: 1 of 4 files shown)


kleinfreund:
Ah, right, very good point. Done.

kleinfreund:
@zachleat Ah, I forgot to mention this. Occassionally, a test fails on Windows. As far as I can tell, the order of the results for `deepEqual` is not stable and might produce results in the wrong order. Probably something to do with async functions? So should some Windows tests break in the future and it makes no sense, you heard it here first: Tests break, but only sometimes. :confused:

zachleat:
Awesome, this is still a good improvement. We can look at the test failures before we add #353.

Issue #104, opened 2018-04-14
Package eleventy into a single binary or executable


chrisdmacrae:
One of the common complaints against almost all other static site generators is installation.

This is especially a problem for Jekyll on Windows.

Zeit/pkg creates application executables of node applications for Darwin, Win32, and Unix systems, and will still allow users to add plugins from the NPM ecosystem.

This should be a late roadmap goal, but I believe it will separate Eleventy from the pack by making it accessible to non-technical users.

This could be deployed to OS-level package managers (e.g, homebrew, package-manager, and apt) in order to make installation and update management a breeze, while still leaving the product as a CLI.

chrisdmacrae:
Note, they are performance considerations here. Packaged node apps may run 1.5x slower than unpackaged apps.

This may not be a problem for eleventy as the major bottleneck is file I/O.

zachleat:
cc @mattdsteele this might alleviate the problem you were going through with requiring Node 6 support

zachleat:
I know Hugo provides binaries and they have been successful there https://github.com/gohugoio/hugo/releases

I like this idea!

zachleat:
https://github.com/nexe/nexe

aral:
Just wanted to add that having a single binary would also make it much easier to include 11ty support in other tools like, *wink, wink* [Site.js](https://sitejs.org) alongside other single-binary static site generators like Hugo and Zola.

PR #391, opened 2019-01-21
Enable packaging with zeit/pkg

buckle2000:
Partially solves #104.

Now, we can run `pkg .` to build binary files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,5 +113,8 @@
       "pre-commit": "lint-staged",
       "pre-push": "npx ava"
     }
+  },
+  "pkg": {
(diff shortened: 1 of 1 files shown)


chrisdmacrae:
For this PR to be accepted we should run the test suite against the src code AND the binary, and develop a formal release process for binaries.  @zachleat

MadeByMike:
Could we add some documentation in this PR to explain how and why people might want to use this?

zachleat:
Ah yeah I don’t really understand what this does—would love some more context here! Please leave a comment and I’ll reopen

PR #421, opened 2019-02-16
Throw an error if includes or data directory is outside of Eleventy’s working directory

### LINKED ISSUE #368: RFC: Restrict Eleventy’s access to directories outside of its realm

kleinfreund:
This pull request is the first step for #368.

- Makes `TemplateConfig` throw an error on construction if either the includes or data directories point outside of Eleventy’s working directory.
- There is also some unrelated clean up in the first commit:

  The `TemplateWriter` class no longer creates its `EleventyFiles` object itself when it is missing. This was only there to ensure that the `EleventyFiles` object is set during testing. Setting it is now a responsibility of the tests itself.

What do you think about the error message or error handling in general?

**Note**: Since there might be users who rely on the current functionality, this is a breaking change.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -268,8 +268,7 @@ class Eleventy {
       this.input,
       this.outputDir,
       formats,
-      this.templateData,
-      this.isPassthroughAll
+      this.templateData
     );
 
     this.writer.setEleventyFiles(this.eleventyFiles);
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -12,7 +12,7 @@ const debug = require("debug")("Eleventy:EleventyFiles");
 // const debugDev = require("debug")("Dev:Eleventy:EleventyFiles");
 
 class EleventyFiles {
-  constructor(input, outputDir, formats, passthroughAll) {
+  constructor(input, outputDir, formats, isPassthroughAll) {
     this.config = config.getConfig();
     this.input = input;
     this.inputDir = TemplatePath.getDir(this.input);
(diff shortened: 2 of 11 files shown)


kleinfreund:
If this looks good to you, I can do the same for the input and output directories.

kleinfreund:
I resolved the conflicts with the latest changes in the main branch; hence, the changes needed to be force-pushed again after rebasing.

d2s:
Good work with the initial concept. This kind of improvements increase my trust in the future of Eleventy. 👍

kleinfreund:
I modified the [first commit](https://github.com/11ty/eleventy/pull/421/commits/530154c32f5847aa0fb9720f31895eccf15f5429) in the following way:

The `TemplateWriter` class no longer creates its `EleventyFiles` object itself when it is missing. This was only there to ensure that the `EleventyFiles` object is set during testing. Setting it is now a responsibility of the tests itself.

I’ll update the original PR message.

Issue #389, opened 2019-01-16
Ignore node_modules if .gitignore exists but is empty


zachleat:
Current behavior: node_modules are ignored by default only if .gitignore does not exist. If a .gitignore file does exist, there is no special exception for node_modules.

Related #383 but that issue is a major version change. This issue is a shorter term improvement that could be minor version implemented.

zachleat:
Thank you @danielstaleiny! Merged!

PR #422, opened 2019-02-16
Check if the .gitignore is empty and if it is ignore node_modules.

danielstaleiny:
This pull request refer to #389


zachleat:
I like where you’re going here. I would prefer if we could get the file contents and trim to length so we could make this work if the file just has whitespace, too?

Also can you change this so it doesn’t modify any of the existing tests? Not sure why the folders needed to be changed there

danielstaleiny:
Main reason for re-ordering the test was that I wanted them to be logically grouped.

danielstaleiny:
I squashed the changes.  all the changes are in the last commit.

zachleat:
Ah ah, I see. Yeah I can see the temptation for that. Modifying existing tests is a big red flag when I’m reviewing things—if you’d like to reorder them I’m happy to accept a new PR that does only this (thank you and sorry 😱)

PR #486, opened 2019-04-11
forces uses of posix path separator in url normalization

### LINKED ISSUE #485: URL filter doesn't work on Windows
When building on Windows and using the URL filter the resulting url ends up with the windows path separator.

This just started recently, so my guess is a dependency was overwriting node's path with upath and has stopped doing that.  

The fix is to use the posix namespace in TemplatePath.normalizeUrlPath,

AndrewAsquith:
This should fix #485


kleinfreund:
I think this is fine. This would probably also fix #353.

zachleat:
Thank you!

Issue #484, opened 2019-04-09
Possible dynamic permalinks bug in 0.8+


robdodson:
Given the following project structure:

```
en/
  en.11tydata.js
  posts/
    foo/
      index.njk
```

```js
// en.11tydata.js

module.exports = {
  language: 'en',
};
```

```
// foo/index.njk

---
layout: post
---
```

```
// _includes/post.njk

---
permalink: /{{language}}/{{page.fileSlug}}/index.html
---
```

The directory data file should inform the permalink in the layout.

Expected behavior:
```
output/en/foo
```

Actual behavior:
```
output/foo
```

This works as expected in `0.7.1` but seems to break in all versions of `0.8.*`.

zachleat:
Hmm… this is surprising to me. The data aggregation code didn’t really change in 0.8.0 and we have tests around inheriting data from layouts.

Also given your repro steps, it seems to work fine:

<img width="664" alt="image" src="https://user-images.githubusercontent.com/39355/55849412-19193000-5b16-11e9-8011-a596fc586eb8.png">

Does the sample project I created work for you? [eleventy-484.zip](https://github.com/11ty/eleventy/files/3061830/eleventy-484.zip)

What environment are you running in?

robdodson:
The zip you sent over works fine, so it must be something in the way my project is setup. I can consistently reproduce it by toggling versions and cleaning out `node_modules`.

If you grab the [v1 branch of our repo](https://github.com/GoogleChrome/web.dev/tree/v1) you might be able to reproduce it. With 0.7.1 installed, folders like `accessible` and `fast` do not appear outside of `en`. When I switch to 0.8.1, they do.

Apologies in advice for the messiness of the project. It's currently midway through migrating from our hand written build scripts over to eleventy. If you take a look all of the relevant bits are in the `src` dir.

zachleat:
I was able to reproduce this, thanks! More info pending!

zachleat:
This was a little bit of a weird one—our refactored logic to find directory names was subtly incorrect! I’ve added a test for your specific case.

The fix for is in 0.8.3.

cc @kleinfreund please review the commit and test (thanks!) https://github.com/11ty/eleventy/commit/f5f5c618920982d52497775740d75cc02d84338c

kleinfreund:
@zachleat I see. Sorry about that. However, I’m afraid both versions allow for another subtle bug. Your initial version (before the refactoring PR) was more correct.

```js
[...]

PR #491, opened 2019-04-14
Fix getAllDirs not handling segments of equal name

kleinfreund:
Reviewing https://github.com/11ty/eleventy/issues/484#issuecomment-482911190. See #484.

The `TemplatePath.getAllDirs` utility was broken by me in #376 and stopped handling paths with multiple segments with the same name correctly. A patched version in https://github.com/11ty/eleventy/commit/f5f5c618920982d52497775740d75cc02d84338c has the same bug, but differently. This pull request should fix variants of the bug.

### DIFF
--- a/src/TemplatePath.js
+++ b/src/TemplatePath.js
@@ -78,9 +78,7 @@ TemplatePath.getAllDirs = function(path) {
 
   return path
     .split("/")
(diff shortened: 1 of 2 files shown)


zachleat:
Awesome, thank you!!

PR #557, opened 2019-06-04
Check for empty tags

### LINKED ISSUE #556: Empty tags in post causes merge failure.
I'm unsure if this is _quite_ a bug, but it's an area where eleventy could have a bit better error handling.

If tags exists in frontmatter but is empty (null?), merging with data files fails.

It would help if eleventy coerced empty tags to `[]` or something.

I think [this section](https://github.com/11ty/eleventy/blob/9378fc3cde4ee866637a1e72de88100f20896c6f/src/TemplateData.js#L340) would want to be edited to test for null. **edit:** pr submitted

## Why I have empty tags:

I create most of my posts from a template file with a load of front-matter prepopulated, like this:

```yaml
---
title: TITLE
subtitle: This is the title
date: YYYY-MM-DD
tags: 
location:
venue:
people:
trip:
series:
primaryImage: path/to/the/image.jpg
---

[[CONTENT HERE]]
```

Up until now, I've *always* had tags on posts. But I'm now moving to deepDataMerge where my one common tag (to set a post type) is in a data file.

It seems the merge of page and data file fails if tags exists but is empty.

You could argue I don't need it above - I find it helpful to remind me what things I might want to add - and I hope an empty item shouldn't cause a crash.

edwardhorsford:
Fixes #556 where empty tags causes the merge to fail.

### DIFF
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -339,8 +339,11 @@ class TemplateData {
 
   static cleanupData(data) {
     if ("tags" in data){
-      if (typeof data.tags === "string" || data.tags === null){
+      if (typeof data.tags === "string"){
         data.tags = data.tags ? [data.tags] : [];
+      }
+      else if (data.tags === null){
+        data.tags = []
       }
     }
 
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -339,11 +339,8 @@ class TemplateData {
 
   static cleanupData(data) {
     if ("tags" in data){
-      if (typeof data.tags === "string"){
+      if (typeof data.tags === "string" || data.tags === null){
         data.tags = data.tags ? [data.tags] : [];
-      }
-      if (data.tags === null){
-        data.tags = []
       }
     }
 
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -338,8 +338,13 @@ class TemplateData {
   }
 
   static cleanupData(data) {
-    if ("tags" in data && typeof data.tags === "string") {
-      data.tags = data.tags ? [data.tags] : [];
+    if ("tags" in data){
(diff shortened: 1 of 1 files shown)


src/TemplateData.js
```diff
@@ -338,8 +338,13 @@ class TemplateData {
   }
 
   static cleanupData(data) {
-    if ("tags" in data && typeof data.tags === "string") {
-      data.tags = data.tags ? [data.tags] : [];
+    if ("tags" in data){
+      if (typeof data.tags === "string"){
+        data.tags = data.tags ? [data.tags] : [];
+      }
+      if (data.tags === null){
```
kleinfreund: This should be an `else if` since the conditions are mutually exclusive. When `typeof data.tags === "string"`, the second condition is evaluated again even though it cannot be true.

edwardhorsford:
@kleinfreund thanks for looking so quickly. I've refactored it a bit so both checks are included. It _sort of_ looked like the ternary check was trying to do this anyway - so I'm using that as well now.

zachleat:
Oh damn @edwardhorsford sorry to do this but I thought your original commit was actually much easier to read. Can you switch back to that and use an `else if` like @kleinfreund suggested? (Sorry!)

edwardhorsford:
@zachleat Updated as requested.

It probably wants tests, but I'm not familiar with adding those, sorry.

I also don't know if there's other stuff that gets merged that should also check for empty.

zachleat:
Thank you @edwardhorsford!!

zachleat:
Just added a test for this in https://github.com/11ty/eleventy/commit/f014f56f26c3df71de3520558c0034e9cae8fb59 if you’d like to have a look @edwardhorsford (it’s one line 😇)

PR #564, opened 2019-06-09
Allow delaying reprocessing on file change.

### LINKED ISSUE #559: Add support for a "cool-down" period before triggering a regeneration.
I have a site which contains multiple uniquely styled sub-pages which each have their own SASS files.  When running eleventy alongside gulp, when I save a single SASS file, gulp regenerates all of the css files which triggers eleventy to regenerate.  Unfortunately, the first file change triggers a regeneration, and the second file change queues up another one.  So as any css file causes eleventy to regenerate everything twice which really slows down iteration.

For example, given this super basic site example:

```
/site
    /events
        /birthday
            index.md
            style.scss
            /media
                thing.ttf, etc.
        /christmas
            index.md
            style.scss
            etc.
```

I have a gulp task that processes all the .scss files to produce `style.css` in the same folder.  The actual sub-sections are relatively complicated but self-contained and they include things like fonts, images, additional sub-pages etc.  Moving all the styles and images to folders outside of the root site would make it difficult to manage each of these sites independently so I would like to keep everything self-contained if possible.  

I would love if there was some sort of configurable delay that would wait after receiving a file modification event before regenerating the site.  The more complex version would be don't regenerate until X ms after the last file change event, so if my gulp task is long running each generated file would keep pushing back the regeneration time a little bit until everything was done.  

This is similar to the [`awaitWriteFinish` property in chokidar](https://github.com/paulmillr/chokidar#performance) which I manually enabled in Eleventy.js and was hoping would work by queuing up file change events quickly enough to cause only a single eleventy rebuild, but no luck.

veleek:
Resolves #559 

Add support for specifying a delay after a file change occurs after which Eleventy will start reprocessing the input files.  As a super simple example, if you have two files open in an editor and you are running `eleventy --serve` and then you execute a save all eleventy will:
* Receive a change notification for the first file.
* Start reprocessing all files.
* Receive a change notification for the second file.
* Queue up a reprocess
* Finish processing all files INCLUDING the already modified second file.
* "Notice" that a file was saved while eleventy was running and then start reprocessing.
This double processing can be time consuming for large projects and is especially noticeable when you have additional external tools running to modify files (like SASS).  If you change a single SASS file, the SASS tool will run and regenerate all CSS files, and the first one to finish will trigger eleventy to rebuild, and then they will immediately re-process a second time after that.  

This PR add support for a new config property `watchTriggerDelay` (open for alternate suggestions).  If `watchTriggerDelay` is provided eleventy will wait until that long AFTER the last file change occurred before it starts reprocessing.  This means that if you set `watchTriggerDelay` to 10 seconds, and save a file, and then wait 5 seconds and save a second file, then it will wait another 10 seconds before starting to process any changes.

As a side-effect of this change, we now also queue up any file changes that do occur while we're processing and will refresh all of them instead of just the very last one.  This change does NOT attempt solve the fact that file changes during processing can result in unexpected output, but it will still trigger an immediate re-process immediately afterwards for any changes that didn't make it into the first "batch".

This change is using an async sleep method by the `es7-sleep` package, but it's an incredibly minimal implementation so if you'd prefer dropping that dependency we can move it directly into eleventy.


zachleat:
Hmm, can we solve this by extending/exposing a configuration API to set chokidar options? In situations like this I’d rather defer to exposing configuration rather than solving it with more code internally.

See:
https://github.com/paulmillr/chokidar#api

Maybe by adding an Object.assign to:
https://github.com/11ty/eleventy/blob/master/src/Eleventy.js#L364

This is similar to what we do with advanced BrowserSync usage here:
https://github.com/11ty/eleventy/blob/master/src/UserConfig.js#L494 and https://github.com/11ty/eleventy/blob/master/src/EleventyServe.js#L63

zachleat:
Actually now that I think about it we may already allow customization of this, because you can set watchOptions in browser sync config: https://browsersync.io/docs/options#option-watchOptions

veleek:
There may have been a bit of a miscommunication in the original issue.  I actually already tried manually overriding the chokidar options on my own build but it doesn't solve the problem.  The chokidar option I referred to (`awaitWriteFinish`) actually just holds off sending file change events until it's sure the file is no longer changing.  It doesn't actually batch up any file change requests so eleventy still receives each one of them separately.  The current behavior looks like this:

Delays for eleventy processing are show way lower than they actually are, but lets just say it's 1000ms.
And lets say it takes gulp about 50ms to process each file.

```
T(ms) |Gulp                 |Chokidar          |Eleventy (event)      |Eleventy processor
0     | Trigger on .scss    |                  |                      |
10    | regenerate all css  |                  |                      |
50    | Generate F1.css --> | Trigger F1 ----->| Trigger Rebuild  --> | Processing F1, F2 and F3
100   | Generate F2.css --> | Trigger F2 ----->| Queue Rebuild        |
150   | Generate F3.css --> | Trigger F3 ----->| Queue Rebuild        |
200   |                     |                  |                      |
250                                                                   |
...                                                                   |
1050                                                                  | Rebuild complete 
-                                                                     | Check if build queued (yup)
1100                                                                  | Processing F1, F2 and F3
...                                                                   |
2100                                                                  | Rebuild complete 
-                                                                     | Check if build queued (nope)
2100                                                                  | Ready to serve files
```

Now assuming we have a 200 ms delay configured.

```
T(ms) |Gulp                 |Chokidar          |Eleventy (event)      |Eleventy processor
0     | Trigger on .scss    |                  |                      |
0     | regenerate all css  |                  |                      |
50    | Generate F1.css --> | Trigger F1 ----->| Queue Rebuild F1 --> | Delay rebuild
100   | Generate F2.css --> | Trigger F2 ----->| Queue Rebuild F2     | Delaying
150   | Generate F3.css --> | Trigger F3 ----->| Queue Rebuild F3     | Delaying
200   |                     |                  |                      | Delayed 200ms.  
250                                                                   | Last file was changed at 150ms.
300                                                                   | Delay is 200ms from then, so 350ms
350                                                                   | Processing F1, F2 and F3
1350                                                                  | Rebuild complete 
1350                                                                  | Check if build queued (nope)
1400                                                                  | Ready to serve files
```

So unfortunately, just passing some configuration to to chokidar doesn't solve the problem.  We explicitly need the ability to delay processing for some period of time.  I  referenced this in the issue, but I think it's worth referring to it here too, but gulp [supports a delay option when watching files](https://gulpjs.com/docs/en/getting-started/watching-files#delay) and it defaults to a non-zero value for the exact same reason we need it here.

Ryuno-Ki:
> Maybe by adding an Object.assign to:
> https://github.com/11ty/eleventy/blob/master/src/Eleventy.js#L364
> 
> This is similar to what we do with advanced BrowserSync usage here:
> https://github.com/11ty/eleventy/blob/master/src/UserConfig.js#L494 and https://github.com/11ty/eleventy/blob/master/src/EleventyServe.js#L63

Those links will be wrong in the future. @zachleat if you are on the site, hit <kbd>y</kbd> to get a canonical URL (which will stay permanent over time). More tips hidden behind <kbd>?</kbd> :-)

veleek:
@zachleat - Any further thoughts on this?  I've been using this for a week now and while it's only one user it seems to be working pretty well for me.

veleek:
Hey @zachleat.  I'm still hoping to get this merged if possible.  Please let me know if there's anything else that you'd like to consider and I'll try to make the change so we can get this closed.

Issue #278, opened 2018-10-17
Slug filter doesn't create url safe slugs.


edwardhorsford:
I've been using the `slug` filter to convert regular strings to strings suitable for use as urls. However I've realised it's not set up to remove apostrophes (and possibly other characters?).

So `it's a test` becomes `it's-a-test`, and when rendered with eleventy the url is `/tags/it&#39;s-a-test/` - the ampersand breaks the url. It also looks ugly!

I see that the [package supports](https://www.npmjs.com/package/slugify#options) supplying a list of characters to remove. 

If the intention is that this filter is used to create 'safe' urls, could Eleventy remove these by default?

---

As an alternative I've added my own filter using the `slugify()` option of the string library. I might stick with this anyway as it seems to do a good job of creating 'pretty' slugs.

hughrun:
I just noticed a similar problem with commas in the title - it leaves the commas in for the automatically slugified permalink.

zachleat:
Oh, hmm. Well that’s not ideal.

I did try `permalink: '/{{ "Hi I''m Zach" | slug }}/'` (note the escaped single quote using `''`) which resulted in the URL `http://localhost:8080/hi-i'm-zach/` which worked in both Chrome and Firefox. Where is the conversion happening in your URL? It doesn’t seem to be coming from the slugify filter.

edwardhorsford:
I think the point is more that I'd expect apostrophes to be be stripped by the filter. And that it would generally make 'nice' urls.

In my case this was most obvious when using automatic anchor links from headings - or from tags from tag pages.

zachleat:
Well as much as I agree with you, this would be a breaking change and would require a major version bump. So I'm trying to figure out the exact impact of this to determine priority. It doesn't seem to be breaking anything in the test case I tried.

edwardhorsford:
Oh totally! Quite a bad breaking change too. Even with a major version bump I suspect you'll want / need to offer a compatibility mode. Although I it only impacts a certain percentage of links, and users will likely have avoided using them, it has the potential to screw up lots of links.

Some thoughts:
* You could expose more of the config of the base package so users can strip these things if they want.
* You could offer a higher level config that users opt in to (and later make default, and offer a 'legacy' option)

FWIW, I think I prefer the `string` library's implementation anyway - to me it handles more edge cases nicer. I'll stick with my own filter anyway.

jevets:
@edwardhorsford 

Out of curiosity:

[...]

PR #572, opened 2019-06-14
Extend Slugify to slug reserved characters

borisschapira:
* Before : `"John's cars" | slug` => `john's-cars`
* After : `"John's cars" | slug` => `johns-cars`

Example Use case: using tags in permalinks, like in https://github.com/11ty/eleventy-base-blog and having a tag that contains an apostrophe.

### DIFF
--- a/src/Filters/Slug.js
+++ b/src/Filters/Slug.js
@@ -1,5 +1,6 @@
 const slugify = require("slugify");
 
+slugify.extend({'\'': '-'})
 module.exports = function(str) {
   return slugify(str, {
     replacement: "-",
--- a/src/Filters/Slug.js
+++ b/src/Filters/Slug.js
@@ -1,9 +1,9 @@
 const slugify = require("slugify");
 
-slugify.extend({'\'': '-'})
 module.exports = function(str) {
   return slugify(str, {
     replacement: "-",
-    lower: true
+    lower: true,
+    remove: /[:\/\?\#\[\]@!$&'()*+,;=]/g
   });
 };


kleinfreund:
Why is that necessary?

borisschapira:
@kleinfreund According to [the documentation](https://www.11ty.io/docs/filters/slug/), the `slug` filter uses the slugify npm package to convert a string into a URL slug.

If your string contains an apostrophe, the apostrophe is not slugified. If you use the so-called `slug` in a permalink, the apostrophe is HTML-encoded instead of URL-encoded. In the example, it becomes `johnd&#39;s-car`, which is invalid for an URL.

kleinfreund:
Oh. Then the apostrophe should not be HTML-encoded in the first place, no?

borisschapira:
Right, but even if it was URL-encoded, the result would not be user and seo friendly: `john%27s-car`. And as slugs are used for a variety of usages, including URLs but also IDs, apostrophes should be encoded too, like other punctuations.

kleinfreund:
Are apostrophes really URL-encoded if they just appear in a URL?

borisschapira:
The apostrophe is HTML-encoded in the generared files: `./_site/tags/john&#39;s-car/index.html`.

Ryuno-Ki:
Plus, search engines will have an easier job without apostrophe: https://support.google.com/webmasters/forum/AAAA2Jdx3sUaKVMfwL6WgE/?hl=en&gpf=%23!topic%2Fwebmasters%2FaKVMfwL6WgE

What about other „reserved characters”? (c.f. Section 2.2 of [RFC 3986](http://www.faqs.org/rfcs/rfc3986.html))

borisschapira:
I tested `:/?#[]@!$&'()*+,;=` and obtained `:@!dollarand'()*+` which means there are 3 groups.

* characters that disappear: `/?#[],;=`
* characters that stay: `@!'()*+`
* characters that are converted: `$` (becomes `dollar`) `&` (becomes `and`)

While I was searching, I found [an interesting note in the doc](https://github.com/simov/slugify/blob/master/README.md#L30), saying that we can remove multiple characters altogether.

I find that quite interesting, as I'm not sure we _should_ replace reserved characters with `-`. What do you thing?

Ryuno-Ki:
Hm, I would like to have more use cases :-)
I mean, what I did in a job was picking a [list of stop words](http://xpo6.com/list-of-english-stop-words/) (for the German language - not available online) and removed them from the URL. For „SEO” (a.k.a. black magic).

But this could mean, the URL structure should be decoupled from the directory one.
I can imagine, offering some hooks for a plugin here might be more worth.

In your case: Replacing reserved characters with dashes and collapse multiple dashes into one would be a good first approach in my opinion.

edwardhorsford:
See #278 for the existing issue on this. I moved to my own filter to do it. Hoping it gets updated at the next major version.

kleinfreund:
> The apostrophe is HTML-encoded in the generared files: `./_site/tags/john&#39;s-car/index.html`.

I consider this a bug. URLs should not contain HTML-entity-encoded strings unless they are actually part of the URL in which case, some characters need to be URL-encoded (e.g. to avoid adding an unwanted `&` character).

I don’t see a reason why perfectly legal characters should be replaced with a delimiter. `/it's-a-test` is a perfectly fine URL path. We just need to take care that characters aren’t HTML-entity-encoded.

borisschapira:
I agree.
However, what's the use of a slug if you can't have a easy-to-read result, or can't use it as an id because of those forbidden characters? Maybe I don't clearly know of to use a slug.

Additional note: if we can't use slug for user-friendly URL, the documentation should be modified.

kleinfreund:
A slug should for a URL should be easier the read the more closely it resembles the input string. Maintaining apostrophes, for example, helps there: I think `it's-a-test` is easier to read than `it-s-a-test`.

That is one use case for transforming an input text into a slugified version. For IDs (e.g. HTML `id` attribute), the criteria is different and has to be considered separately.

borisschapira:
I understand. So maybe my PR is useless, I'll develop a plugin. Thanks for you time.

edwardhorsford:
I agree with @borisschapira - I see the intention of the filter as wider than making a "valid" url - I expect it to produce pretty urls. Take arbitrary string input, output a sensible url. This is how many other static site generators do it.

@borisschapira see my other issue for the filter I use to produce 'pretty' urls.

borisschapira:
@edwardhorsford I like the idea of extending Eleventy when I need to so I made [a plugin](https://github.com/borisschapira/eleventy-plugin-better-slug) for that.

Peggy535:
@borisschapira - A huge thanks for your efforts with your plugin! It works a dream. Sincere appreciation and thanks. 

Wayne Pegg

borisschapira:
@Peggy535 Oh thanks for the feedback! I'm glad it works as you expected!

PR #582, opened 2019-06-20
Add JSDoc comments

Ryuno-Ki:
Hi @zachleat,

just want to get your feedback on this before I invest too much time into it.
Would you accept JSDoc comments in your code?

This way, editors would be able to suggest Autocomplete, you could use them for [typing with flow](https://www.npmjs.com/package/flow-jsdoc) and you can render documentation for the implementation.

I can't speak for others, but I find it sometimes difficult to understand, what's going on.

Since I document my Node.js projects, I developed some opinions.
For example, using [ink-docstrap](https://www.npmjs.com/package/ink-docstrap) for rendering more appealing JSDoc html files.

Read more about [JSDoc](https://jsdoc.app/index.html) online (old URL was usejsdoc).


kleinfreund:
Yes, I assume that's welcome. I documented code in my contributions with JSDoc, too.

Ryuno-Ki:
Okay, since I got a positive comment + a :+1:, I added docstrings to the rest of `Eleventy` class.
More will follow after this one get merged in.

Would help if somebody could update the _tbd._s after a merge :sweat_smile: 
Wasn't completely sure, what the code is supposed to do (another positive side effect when adding documentation: it forces you to write cleaner code :-D )

src/Eleventy.js
```diff
@@ -11,33 +11,82 @@ const config = require("./Config");
...
 
+/**
+ * @module @11ty/eleventy/Eleventy
+ */
+
+/**
+ * Runtime of eleventy.
+ *
+ * @param {String} input - Where to read files from.
+ * @param {Output} output - Where to write rendered files to.
```
Munter:
```suggestion
 * @param {String} output - Where to write rendered files to.
```

Ryuno-Ki:
@zachleat Is there something blocking this from being merged?

zachleat:
Thanks for this @Ryuno-Ki!

PR #592, opened 2019-06-28
chore(Eleventy): Reordered the methods of Eleventy class.

Ryuno-Ki:

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -48,6 +48,10 @@ class Eleventy {
     return dir;
   }
 
+  get watcherBench() {
+    return bench.get("Watcher");
+  }
+
   setDryRun(isDryRun) {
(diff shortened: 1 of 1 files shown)


Ryuno-Ki:
So basically:
* Constructor
* Getters
* „Setters”
* „Getters”
* Sync methods
* Async methods
* Private methods

Coming from a Python background, this is an order which makes it somewhat easier to navigate.
Do you agree?

zachleat:
Hmm—I can’t say that I do agree here no—sorry. Maybe the biggest sticking point for me is separating sync/async/private methods? I try to group these by how they’re used.

I will say that reordering PRs are a tough sell at this point. 🤷‍♂️ Not to say there isn’t value here but they’re just really low priority and require more review work than they’re worth imo. Super sorry 😢

PR #612, opened 2019-07-15
Throw when localConfig is a function that returns a promise

Munter:
I attempted to use an async function in `.eleventy.js` because I wanted to try something depending on disk access, and usually do those things async. It turned out everything I tried to configure didn't work. The reason for this was that the config resolution doesn't handle async or promises, and just went ahead and merged my returned promise with the default config.

This PR makes the config resolution throw if a localConfig function returns a promise. Changing the entire flow to support async can be a major undertaking, but at least this addition will help future users identify their mistake clearly if they make the same one.


zachleat:
Beautifully concise PR, thank you.

PR #627, opened 2019-07-25
Improve JavaScript templates

### LINKED ISSUE #622: Support class properties in js templates
```js
module.exports = class {
  data = {
    title: 'hi',
  };

  render(data) {
    return `
      <p>${data.title}</p>
    `;
  }
};
```

Class properties are a relatively new JS feature, but are supported in recent-ish versions of Node.

I would expect the above to render `<p>hi</p>`, but it isn't picking up the data.

zachleat:
* When using class-based templates, use a single object instance for data and rendering.
* Add support for object-based JS templates a la `module.exports = { data: function() {}, render: function() {} }` and `module.exports.data = function() {}; module.exports.render = function() {};`
* Add support for Node v12+ feature: class fields a la:

```
module.exports = class {
  data = {
    title: 'hi',
  };

  render(data) {
    return `
      <p>${data.title}</p>
    `;
  }
};
```

* Fails gracefully when class or object definition is missing a render method (returns `""`), behaves similar to a data file.

Fixes #622.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - 8
   - 10
+  - 12
 before_script:
   - npm install
 script: npm run test
--- a/src/Engines/JavaScript.js
+++ b/src/Engines/JavaScript.js
@@ -15,22 +15,33 @@ class JavaScript extends TemplateEngine {
     return result;
   }
 
+  _getInstance(mod) {
+    if (typeof mod === "string" || mod instanceof Buffer || mod.then) {
+      return { render: () => mod };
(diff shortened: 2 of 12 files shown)


zachleat:
Okay I think this is good to merge after Travis does it’s thing.

jakearchibald:
Yay! Glad I was useful

reubenlillie:
@jakearchibald, thank you for your work here. @zachleat, thanks for receiving it. Class syntax in JS feels so foreign. This added flexibility makes teaching 11ty to beginners so much easier.

PR #635, opened 2019-07-28
User data hooks

### LINKED ISSUE #630: Hooks to allow plugins to customize data file formats (think: yaml, toml, etc)
Similar to #117 but for data files instead of template files.

Related #265.

gloryofrobots:
This PR resolves #630.

User API looks like 
```
  eleventyConfig.addDataExtension("yaml", contents => yaml.safeLoad(contents));
};
```
I am using Map for storing pairs {extension:parser}. 
**Override order**: in the case of 
```
  eleventyConfig.addDataExtension("yaml", contents => yaml.safeLoad(contents));
  eleventyConfig.addDataExtension("json2", contents => JSON.parse(contents));
```
json2 files will override data from yaml in the case of duplicate keys.

For tests, I created separate file UserDataExtensionsTest.js and test stubs (test/stubs-630).

I have three questions:

- In TemplateData.getDataValue was a line ` if (ignoreProcessing  || extension === "js") ` which I rewrote as  
``` if ((ignoreProcessing && extension === "json") || extension === "js")``` for my code to work. Can we remove this ignoreProcessing from here?  The block which manages JSON and custom data files has separate logic for ignoreProcessing already.
- In the tests, I attached Map with data formats to the TemplateData in the very ugly fashion.
```
function injectDataExtensions(dataObj) {
  dataObj.config.dataExtensions = new Map([
    ["yaml", s => yaml.safeLoad(s)],
    ["nosj", JSON.parse]
  ]);
}
```
 Is there a more elegant way to do it? Preferably I would like to use the user API and not `TemplateData._setConfig`.
- I found that global data from _data dir does not have data priority order, because of glob usage.
Does it need fixing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "toml": "^3.0.0",
     "viperhtml": "^2.17.0",
     "vue": "^2.6.10",
-    "vue-server-renderer": "^2.6.10"
+    "vue-server-renderer": "^2.6.10",
+    "js-yaml": "^3.13.1"
   },
   "dependencies": {
     "browser-sync": "^2.26.7",
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -110,10 +110,19 @@ class TemplateData {
 
   async getTemplateDataFileGlob() {
     let dir = await this.getInputDir();
-    return TemplatePath.addLeadingDotSlashArray([
+    let paths = [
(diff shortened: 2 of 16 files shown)


zachleat:
Filed #794 to handle the global override question separately.

zachleat:
Merged your comments with my changes to a new PR: #798. Closing this one.

PR #650, opened 2019-08-08
Fix broken error reporting when writing templates

peterjanes:
Adds missed `new` operators from commit 320af9bb0206d2befb0eafd4007481e3965a28ec


zachleat:
Thanks!

Issue #143, opened 2018-06-21
Add twig templates


zachleat:
https://github.com/twigjs/twig.js
https://github.com/ericmorand/twing

Relevant: https://twitter.com/zachleat/status/1009783307364720640

Twig is popular with PHP devs, I think? Wordpress/Drupal/etc

alexcarpenter:
This would be super helpful! I have been wanting generate a static styleguide for a Craft CMS site that relies on Twig.

fvsch:
I like Twig a lot, but JS implementations are not that great (at least Twigjs was not in my experience, it might have gotten better and Twing might be great).

Also when you have Nunjucks I’m not sure Twig is needed at all. (Both Nunjucks and Twig are inspired by jinja2.)

zomars:
Twing > Twig.js any day!

ericmorand:
I'd like to contribute on this.

iparr:
Like @alexcarpenter - I would love to use 11ty for a Craft CMS styleguide / design system. I also build out static sites using https://github.com/zimmen/gulp-twig . I like to use Twig because I used Twig in a corporate setting and still stick with it when I can for the smaller projects I work on now.

As a trial of 11ty, I tried to convert an existing Twig project into Nunjucks. It's great, but support for things like checks for iterable variables, passing variables using with, all the shorthand if? stuff, using a block more than once in a template - and a bunch of more powerful stuff I'd probably discover if I tried to convert a more complex project - is all dearly missed.

11ty seems very flexible - all my other bits and bobs fell into place - so this is the one issue holding me back. I'll still use it on a fresh project, but I already miss using it on some of my other work.

(I don't want to devalue the work that has gone into this project thus far. (Was invited to contribute my thoughts)[https://twitter.com/eleven_ty/status/1094697181024579584])

notpushkin:
I've thought Twig templates are rather easy to port to Nunjucks – are there any critical differences?

SimonEast:
@notpushkin - they do appear very similar on the surface, but there are some differences when you get into details. One example I encountered today was that in Twig you can pass variables into an included file, whereas in Nunjucks the include just uses the scope of the parent. iparr also mentioned some differences above.

It seems like Twig is a more mature language compared to Nunjucks.

zachleat:
It lives! https://www.npmjs.com/package/@factorial/eleventy-plugin-twig

iparr:
Can't wait to take this for a spin. Nice one.

[...]

Issue #117, opened 2018-05-10
Custom File Extension Handlers


zachleat:
Extension to the configuration API that would allow developers to add their own code to handle a new file extension template type.

Some fake code to get the discussion started:

```
const Vue = require("vue");
const renderer = require("vue-server-renderer").createRenderer();

eleventyConfig.addExtension("vue", function(path, fileContents) {
  return function(data) {
    var app = new Vue({
      template: fileContents,
      data: function() {
        return { data };
      }
    });

    return renderer.renderToString(app);
  };
});
```

Will process all `*.vue` files in the `--input` directory and write the files to `--output`.

Needs to:
* Return a repeatable compiled template function (accepts `data` object)

paulrobertlloyd:
Would this feature I could use the existing Markdown template renderer, but with `.markdown` as the recognised extension instead of `.md`? In which case, 👍.

zachleat:
Hmmmmmmmm. Yeah this could likely be extended to do extension aliasing. I reserve the right to create a new issue for this later though 😇

zachleat:
@munter shared tolk https://www.npmjs.com/package/tolk with me which might be very easy very useful crossover with this issue.

https://twitter.com/_munter_/status/1092824532832923653 with

zachleat:
As exposed in #464, some might want the option to override current implemented file extensions with this method too.

shreyasminocha:
Is this on the agenda yet? Seems to be the top issue in `needs-votes` by a decent margin.

pspeter3:
Could you use https://www.npmjs.com/package/rechoir or https://www.npmjs.com/package/interpret to handle this?

PandawanFr:
Any updates on this? Would love to be able to minify my CSS and other files without any hacks.

cfjedimaster:
Same - came here hoping to find an easy answer. :) I tried `setLibrary` cuz it sure looked like it would work, since it lets you override liquid for example, but it didn't work.

BrentARitchie:
Just wanted to add a quick use case. I sometimes rename my JS and CSS files to .liquid or .njk to do some template processing on them, for instance to add a list of is files to async load or to fill in special classes in CSS files or to use special shortcodes for something like mixins. I have also setup transforms for things like minifying HTML, CSS, etc...

[...]

PR #656, opened 2019-08-14
Add Twig support

pieterbeulque:
### Background

I want to use Eleventy as much as possible, but often have to migrate work later to either Craft or WordPress (using Timber), which means having to convert everything to Twig.

Nunjucks was an option but didn't cover everything, so here we are 😁 

### Status

I copied over all Nunjucks-related code and am now trying to make all tests pass as a start. I guess we'll go from there and keep this PR open for status updates.

Feel free to collaborate.

### Related

See #143


pieterbeulque:
@zachleat - How important is it to match templating functionality across all templating languages?

I found this table as an overview of possible Twig extensions and think it'd be best to stick to what the language supports, instead of doing tricks like [`10da170`](https://github.com/11ty/eleventy/pull/656/commits/10da170d350d0ff27f7d51337474b1ac6e51a667).

```========== ========================== ========== =========================
What?      Implementation difficulty? How often? When?
========== ========================== ========== =========================
*macro*    trivial                    frequent   Content generation
*global*   trivial                    frequent   Helper object
*function* trivial                    frequent   Content generation
*filter*   trivial                    frequent   Value transformation
*tag*      complex                    rare       DSL language construct
*test*     trivial                    rare       Boolean decision
*operator* trivial                    rare       Values transformation
========== ========================== ========== =========================
```

pieterbeulque:
🎉 I just went through the whole _Getting started_ section of the Eleventy-documentation with Twig templates and everything works! 🎉

pieterbeulque:
This is ready (with minimum functionality), but I can't get relative includes to work because of how Twig works with the filesystem. Any pointers on how this was fixed for e.g. Nunjucks (if that works the same) would be very welcome.

tylersticka:
This progress is exciting! It would make transitioning from or sharing resources with [Timber](https://www.upstatement.com/timber/) _way_ easier.

> This is ready (with minimum functionality), but I can't get relative includes to work because of how Twig works with the filesystem.

I'm a bit stumped as well, as you're already using [`TwingLoaderChain`](https://ericmorand.github.io/twing/api.html#twingloaderchain) with both a relative and absolute loader (which is what I would have tried). 🤔

pieterbeulque:
> It would make transitioning from or sharing resources with Timber way easier.

That's my use case as well! My CMS-backed work is either in Craft CMS (Twig) or WordPress (Twig through Timber).

> you're already using TwingLoaderChain with both a relative and absolute loader (which is what I would have tried).

Yeah, I thought that'd be a fix, but how I understand it, Eleventy grabs a file's contents and parses it (to handle frontmatter etc) and passes that contents as a string to the templating engine. So you're in fact always falling back to the StringLoader, that has no notion of where the sourcefile lived on the filesystem. That's where I'm stuck now.

I got to a point where I got it working from the filesystem, and includes worked, but then it didn't support any of the other Eleventy features like frontmatter etc.

pieterbeulque:
🎉 I managed to move an [existing Eleventy-based website](https://mementowoordfestival.be) that was using Nunjucks to Twig successfully!

Two things threw errors compared to Nunjucks:

1. `foo|safe` had to change to `foo|raw`
2. `foo|replace('foo', bar')` has to change to `foo|replace({ 'foo': 'bar' })`

But I guess that means this works as a basic implementation and might be worth merging upstream, regardless of the known issues with relative imports and other (more advanced) Eleventy features?

zomars:
Is something preventing this PR to be merged?  I would love to finally use twig with 11ty! 🙌

pieterbeulque:
Hi @zomars, it will probably have lots of oddities and bugs, but as a first version I think it's ready to be used in the wild.

zachleat:
I really love this work! However my hunch is that this should go after #117 as a plugin. Really trying to avoid more languages in core right now, unless they’re super popular, so I don’t get too overloaded here. Happy to leave this open and continue the discussion though

pieterbeulque:
@zachleat Totally get that, I agree that core shouldn't try to support everything 🙂I'm cool with moving this into some kind of plugin

zomars:
Or maybe we could add this last one? 😅

zachleat:
Going to close this for now—sorry! Please follow along at #117!

cbirdsong:
Has this been moved to a plugin anywhere?

zachleat:
It lives! https://www.npmjs.com/package/@factorial/eleventy-plugin-twig Also note that #143 is the home base for this one

pieterbeulque:
Let's go! Thanks to everyone involved! 🚀

Issue #596, opened 2019-07-04
11ty.js templates not updating on --watch/serve (Windows, at least)


hcavalieri:
**Describe the bug:**

Templates written in Javascript are properly watched by Eleventy, but changes made to them don't propagate to resulting `_site`

**To Reproduce**

1. Create an empty directory and include a simple `index.11ty.js` template there;
2. Run `eleventy --serve` in the dir
3. Make changes to `index.11ty.js` and see if they're reflected in the `_site` folder

**Expected behavior**

Say I were to change my layout from `'<h1>Home here</h1>'` to `'Home'`, I'd expect `_site/index.html` to become solely `home`.

**Screenshots**

I've actually recorded a video of this behavior [here](http://youtu.be/xroXjwbGZMg?hd=1).

**Environment:**

 - OS and Version: Windows 10
 - Eleventy Version `0.8.3`

**Additional context**

For now I've been simply `eleventy build`ing every time I make a change, but you can imagine how unproductive and upsetting this is... 😔 Oh, and the other templating languages are not an  option for my current project.

zachleat:
Great video and description, thanks!

I’m almost certain you found a bug.

I’d bet this has something to do with the pieces of our code that reset the Node require cache. I’m guessing there is a bug with how this is interpreted in Windows. If you want to start investigating (depending on your familiarity with Node) you can find them by searching the code-base for `delete require.cache`. If you want to add some console.logs there and see what outputs, that’d be helpful!

hcavalieri:
Alright, not sure I'm the best at the bugging, but here we go:

1. I got into `src/Engines/Javascript.js` an went into the function `initRequireCache`, which has a line `delete require.cache[requirePath]`
2. Then I added the following before the if statement: `console.log(inputPath, requirePath, requirePath in require.cache)`
3. And got the following when I saved my minimal `index.11ty.js` file: `./index.11ty.js C:/desenvolvimento/tests/11ty-debug/index.11ty.js false`
4. And I tried re-running the process, but `requirePath in require.cache` is always false.

So apparently the cache is never deleted, but that's as far as I understand from this haha

Any other debugging you'd like me to do?

I'm actually going to test this on Glitch.com to see how this behaves and then get back to you 😉

cschuller:
The cache uses the real location of the module, to delete it, I recommend to `resolve` it first.

`delete require.cache[require.resolve(fullModuleName)];`

https://nodejs.org/api/modules.html#modules_require_resolve_request_options

[...]

PR #683, opened 2019-09-05
Fix require cache cleanup on Windows (#596)

cschuller:


zachleat:
Hot damn, that’s a clean PR—thanks!

PR #686, opened 2019-09-07
Line endings

cschuller:
The motivation behind this PR is to improve development on different operating systems.

Adds .gitattribute file based on the common git attributes file provided 
[here](https://github.com/alexkaratarakis/gitattributes/blob/master/Common.gitattributes)

Uses the MIT license, so good to go. ;-)

Added all file extensions that are used in this repository.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,195 @@
+# GITATTRIBUTES
+## Based on https://github.com/alexkaratarakis/gitattributes/blob/master/Common.gitattributes (MIT LICENSE)
+
+# Auto detect
+##   Handle line endings automatically for files detected as
+##   text and leave all files detected as binary untouched.
+##   This will handle all files NOT defined below.
+*                 text=auto
+
+# Source code
+*.bash            text eol=lf
+*.bat             text eol=crlf
+*.cmd             text eol=crlf
+*.coffee          text
+*.css             text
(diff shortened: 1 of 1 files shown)


zachleat:
Hey @cschuller: what problem does this solve?

cschuller:
Hello @zachleat 

When using git for Windows, the default (global) configuration uses `\r\n` for check-out and commit with `\n`. 

Running `eslint` will then result in 300+ errors, because of the `linebreak-style` rule. 

Options to improve the developer experience:
1) Some documentation somewhere (and let the developer configure git)
2) Disable the `linebreak-style` rule (and hope for the best)
3) Use a repository related git attributes file to configure git to use `\n` for checkout and commit.

Whenever I clone a repository I do not want to think about white-spaces and line endings (boring stuff) so I like to use .editorconfig and git attributes.  

By the way sorry that I missed the prettier stuff.

Ryuno-Ki:
+1 for `.editorconfig`.
Haven't played around with `.gitattributes` (but configured my git globally).

zachleat:
Sorry for pushing back here—what’s the impetus for running eslint on the Eleventy source?

mvolkmann:
I would like to run ESLint along with Prettier on any .js files in an 11ty project.

PR #688, opened 2019-09-07
Fixes some eslint errors

cschuller:


src/TemplateLayoutPathResolver.js
```diff
@@ -70,9 +69,7 @@ class TemplateLayoutPathResolver {
   getFileName() {
     if (!this.filename) {
       throw new Error(
-        `You’re trying to use a layout that does not exist: ${
-          this.originalPath
-        } (${this.filename})`
+        `You’re trying to use a layout that does not exist: ${this.originalPath} (${this.filename})`
```
zachleat: This is a prettier format

src/TemplateLayoutPathResolver.js
```diff
@@ -82,9 +79,7 @@ class TemplateLayoutPathResolver {
   getFullPath() {
     if (!this.filename) {
       throw new Error(
-        `You’re trying to use a layout that does not exist: ${
-          this.originalPath
-        } (${this.filename})`
+        `You’re trying to use a layout that does not exist: ${this.originalPath} (${this.filename})`
```
zachleat: This is a prettier format

src/TemplateMap.js
```diff
@@ -509,18 +509,13 @@ class TemplateMap {
...
-            page.outputPath
-          }\`. Use distinct \`permalink\` values to resolve this conflict.
-  1. ${entry.inputPath}
-${permalinks[page.url]
-  .map(function(inputPath, index) {
-    return `  ${index + 2}. ${inputPath}\n`;
-  })
-  .join("")}
-`;
+          warnings[page.outputPath] = [
```
zachleat: This is a prettier format

src/TemplateLayoutPathResolver.js
cschuller: Hello @zachleat 

I am using the local installed prettier 1.18.2 and running it will not change this line.
I also checked against the online [playground ](https://prettier.io/playground/) - no changes.

When I use line `85-87` then prettier will change it to `82`

Do I miss a prettier configuration somewhere?
Should we add a `.prettierrc` config file? 

What can I do to reproduce your preferred style?

zachleat:
Hey @cschuller sorry for the super late response here—is the issue here that Prettier and ESLint are just disagreeing on a few things?

Prettier is run using `lint-staged` here: https://github.com/11ty/eleventy/blob/0c130ebe50f345c473e3fc0d990ef60fc2597266/package.json#L73

PR #696, opened 2019-09-17
Allow delaying reprocessing on file change.

### LINKED ISSUE #559: Add support for a "cool-down" period before triggering a regeneration.

veleek:
Resolves #559

This is a re-creation of PR #564, updated for version 0.9.0 which introduced some changes in the file watch logic.  This still works the same as before (and was actually a little easier to implement), but I'm not familiar with how the incremental file modification logic is done so that should be verified.


src/Eleventy.js
```diff
@@ -391,39 +392,68 @@ Arguments:
...
-      path &&
-      TemplatePath.startsWithSubPath(path, this.eleventyFiles.getIncludesDir());
+    if (this.config.watchTriggerDelay && !immediate) {
+      let timeSinceLastTrigger = new Date() - this.lastTriggerTime;
+
+      while (timeSinceLastTrigger < this.config.watchTriggerDelay) {
+        let triggerDelaySleep =
+          this.config.watchTriggerDelay - timeSinceLastTrigger;
+        console.log(`Waiting ${triggerDelaySleep} ms before processing.`);
+        await sleep(triggerDelaySleep);
```
zachleat: Is it possible to do this with a `setTimeout` instead? Might save a dependency.

veleek: Yup, I can totally do that. I made a comment about this on the original PR since I think the pulled in library is literally 5 lines of code so it seemed like a bit of overkill but wasn’t sure what your preferences would be.

veleek: Yup, I can totally do that. I made a comment about this on the original PR since I think the pulled in library is literally 5 lines of code so it seemed like a bit of overkill but wasn’t sure what your preferences would be.

zachleat:
I think I genuinely owe you an apology here. I didn’t get this merged in and it overlapped with work I was doing for https://github.com/11ty/eleventy/issues/977

I didn’t remember that it existed until I got about 90% of the code in and remembered theis PR. I came back to check on what you did here and it matched up so closely to what I implemented. However, I did not yet have the specific thing you asked for: the watch trigger delay. But I did use some of your ideas and worked on the remaining implementation to solve #559.

Sorry I didn’t get to this in time. Completely my fault. It was good code and I appreciate the work/ideas. I am going to close this PR as unmerged but know that it did help with the implementation that is shipping with 0.11.0. And the feature for #559 will ship with 0.11.0.

Thanks—and sorry again! ❤️

veleek:
Don't worry about it!  I know how hard it can be to find time to work on open source projects.  Eleventy worked great for me, and I'll continue to look for improvements I can make and features I want to ad.  Thanks for the hard work!

Issue #677, opened 2019-09-03
.gitignore with any "!" prefix item causes eleventy to not process any files


bfelda:
**Describe the bug**
If your gitignore contains "!" prefixed items, such as the default VSCode items, eleventy will fail to process any files.  

.gitignore
```
# IDE - VSCode
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
.history/*
```

**To Reproduce**
1. Add the above to your .gitignore file
2. run eleventy
3. note "Processed 0 files in x seconds"
4. note _site dir is added but is empty

**Expected behavior**
files get processed by running eleventy

**Screenshots**

**Environment:**
 - windows 10
 - Eleventy Version 0.9.0

**Additional context**

ogdenstudios:
Confirmed with the example `.gitignore` on OSX 10.14.6 and Eleventy 0.9.0. 

This doesn't seem to be an issue with Eleventy 0.8.3

peter-sharp:
Confirmed on Windows 10 with the following .gitignore using Eleventy 0.9.0:
```
node_modules
.env
_site/*
!_site/assets
```

Ryuno-Ki:
As workaround you can copy your `.gitignore` into an `.eleventyignore`, remove the lines starting with the exclamation mark and [tell eleventy to not look into the gitignore](https://www.11ty.io/docs/ignores/#opt-out-of-using-.gitignore).

zygimantas:
I was also hit by this issue.
Additionally, I have noticed that with such config:
```
const input = "./src/Website/Areas/XXXXXAmp";

config.setUseGitIgnore(false);
config.addPassthroughCopy(`${input}/resources`);
config.addPassthroughCopy(`${input}/robots.txt`);
config.addPassthroughCopy(`${input}/favicon.ico`);
```

where `/Amp` directory exists and `/XXXXXAmp` does not, it fails, as expected in 0.8.3 with TemplatePassthroughManagerCopyError, but in 0.9.0, the output is:

`Copied 3 items and Processed 0 files in 0.02 seconds (v0.9.0)`

Of course, no files are copied, because `/XXXXXAmp` does not exist.

philiprenich:
Using the `DEBUG` mode shows that Eleventy fails on the doubled `!`. I'm shown this string in output `!./!.vscode` where I had `!.vscode/` in the .gitignore

zachleat:
Fixed by #709.

PR #709, opened 2019-09-23
Resolves #693 and #677

### LINKED ISSUE #693: .eleventyignore doesn't support negative patterns
**Describe the bug**
`.eleventyignore` files that use negative ignore patterns return unexpected results because eleventy modifies the paths a bit.

**To Reproduce**
1. Create an eleventy ignore file with a regular pattern and a negative pattern
    ```
    site/**/*
    !site/events/2018/**
    ```
2. Run eleventy ignore
3. Note that no files are copied

**Expected behavior**
Negative patterns should allow you to re-include ignored paths

**Environment:**
 - OS and Version: Windows
 - Eleventy Version: 0.9.0

**Additional context**
Eleventy modifies the patterns in the ignore file by prefixing a relative path header (`./`) and then adding the `!` to the beginning.  As a result, the example above (`!site/events/2018/**`) gets turned into `!./!site/events/2018/**` when being passed to fast-glob.  Fast glob processes negative patterns separately so it pulls that pattern out and removes the ! and actually strips the ./ before using it, so it ends up using `!site/events/2018/**` as the negative pattern and this basically says ignore anything that's not a match for `site/events/2018`.  Since `site/events` is not a match for this, the whole folder gets excluded and fast-glob never scans any deeper for files that SHOULD be included.

I haven't had a chance to try it out yet, but this problem could potentially be solve by not modifying the patterns in `.eleventyignore` and using the new `ignore` parameter to fast-glob which handles the negation for you.  I haven't looked too closely at what modifications eleventy does to see if they're necessary for some of the output processing so hopefully it's just not necessary.

Note: This is a new issue related to the upgrade to fast-glob 3.X.  Definitely prefer to use the new version, but would love to have it continue to use .eleventyignore the same way I was before.

micahmills:
Checking for negative patterns in gitignore and eleventyignore files and and not processing them as files to ignore. This allows you to use the negative pattern (!filename) in your .gitignore and .eleventyignore files without eleventy skipping over everything and not processing your files.

### DIFF
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -183,7 +183,9 @@ class EleventyFiles {
         })
         .filter(line => {
           // empty lines or comments get filtered out
-          return line.length > 0 && line.charAt(0) !== "#";
+          return (
+            line.length > 0 && line.charAt(0) !== "#" && line.charAt(0) !== "!"
+          );
         })
         .map(line => {
           let path = TemplateGlob.normalizePath(dir, "/", line);


micahmills:
While I have used open source software for a long time, I am pretty new to contributing. So if I am not 100% sure I did the whole pull request properly. Please let me know if I should do something differently.

Ryuno-Ki:
From what I can tell, it looks okay.
Keep in mind, that this is developed in peoples' spare time.
Be patient :-)

zachleat:
I think there is a bug here, yes. But there are two parts of it, right?

1. Get Eleventy not to ignore everything, which this does :)
2. Support `!` patterns in `.gitignore` and `.eleventyignore` to allow entries to “punch through” the ignore globs, which this doesn’t do right?

I’m okay to merge this and file the other as a separate thing, I just want to make sure.

e.g. I have a `.gitignore` that looks like:

```
./test/**
!./test/test.md
```

I would expect Eleventy to process `test/test.md` but nothing else in the test folder.

zachleat:
Filed #804 as a follow up to the second part here.

zachleat:
I lied, going to use the pre-existing #693 as the follow up instead.

micahmills:
You are correct this PR only fixes the first part of the problem, but not the second. If I get some time today I will try to work on the second part.

I saw in your comment on that 0.10.0 will include a debug warning for this. I think that will be a huge help. It took me hours to sort out what exactly was happening at first. 

Thanks for you work on this!

Issue #708, opened 2019-09-23
Handling multiple tags in pagination collection


saenglert:
Hello everyone,
I'm trying to work with Pagination and I would like to pass a collection of data that contains multiple tags.

Example:
```
pagination:
    data: collections.news
```
Takes every item with the tag `new`.

What I would like to do is something akin to:
```
pagination:
    data: collections.getByTagNames("news", "domestic")
```
Which would return all items that have both the `news` and `domestic` tag.

I know that I can add custom collections via the config or add a custom filter. #608

However, both solutions are somewhat impractical to me:
 * Custom Collection: If another tag gets added I need to also add a new collection that includes this tag and also add every possible combination collection of  the tags for filtering.
* Custom Filter: As `data` is a string I can't make function calls at this point in frontmatter and if I use the filter later in the template pagination is off because the pagination result and filter result possibly don't match any more. 

I've noticed #397 proposed (to my understanding) exactly this earlier this year. Has anything come of that?

zachleat:
We can close this out because #711 was merged, yeah?

zachleat:
Closing, this will be available in 0.10.0-beta.2

PR #711, opened 2019-09-26
Add getFilteredByTags method and tests

saenglert:
Hello everyone,
in the light of #708 I decided to go with the custom collections for now while my amount of tags is still manageable. However, when creating my custom collections I discovered that there is no filter collections by multiple tags either and the `getFilteredByTag` method is not chainable. So I implemented a method that accepts either an array of values to filter by or a single value:

Single value:
Call is forwarded to the existing `getFilteredByTag` method and its result returned. Strictly speaking the method should return an error at this point and but hey this is JS and I'm a generous god ... err developer (_\*insert Xerxes meme\*_).

Array of values:
The method returns an array of elements which contain every value from the input array in its list of tags.

There are currently still errors in the tests but I'm not sure where they are coming from so input is highly appreciated.

Cheers

Edit: Almost forgot to mention, I also trimmed the `getFilteredByTag` methods a bit by switching to array methods instead of using manual comparisons.

### DIFF
--- a/src/TemplateCollection.js
+++ b/src/TemplateCollection.js
@@ -60,10 +60,7 @@ class TemplateCollection extends Sortable {
     });
   }
 
-  getFilteredByTags(tags) {
-    if (!Array.isArray(tags)) {
-      return this.getFilteredByTag(tags);
-    }
+  getFilteredByTags(...tags) {
     return this.getAllSorted().filter(item =>
       tags.every(requiredTag => {
         const itemTags = item.data.tags;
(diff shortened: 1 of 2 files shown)


saenglert:
Apparently the tests do fail locally but not in travis-ci?!

saenglert:
During lunch break I realized I could get around this whole "single value vs array of values" thing by using [es6 rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)

zachleat:
Thanks, this looks great!

PR #718, opened 2019-10-01
Template utility cleanup

kleinfreund:
This is a maintenance pull request to fix some linter errors and streamline some Eleventy utilities.

- Turns `TemplatePath` utility into a class with static methods to match other utility classes in the Eleventy code.
- Removes the `TemplateGlob.map` utility in favor of calling its underlying methods. There seems to be no need to wrap a `Array.prototype.map` call like it is done currently. Tests for the removed method were removed.
- `Eleventy.setWatchTargets`: Adds a parameter type annotation to `Eleventy.setWatchTargets` to fix linter errors.
- `Eleventy.logFinished`: Adds appropriate casts to explicit number types in mathematical operations to silence linter errors. These conversions do happen upon calling operators like `-` anyway.
- `TemplateMap.checkForDuplicatePermalinks`: A bit of moving around code to make the output generation more readable. Not entirely happy with the result. We can tweak or discard this change.
- `Template.getTemplateMapEntries`: Add missing `_pages` property to the `entries` array to silence linter errors.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -240,12 +240,12 @@ class Eleventy {
     );
 
     let versionStr = `v${pkg.version}`;
-    let time = ((new Date() - this.start) / 1000).toFixed(2);
+    let time = ((new Date().getTime() - this.start.getTime()) / 1000).toFixed(2);
     ret.push(`in ${time} ${simplePlural(time, "second", "seconds")}`);
 
     if (writeCount >= 10) {
       ret.push(
-        `(${((time * 1000) / writeCount).toFixed(1)}ms each, ${versionStr})`
+        `(${((Number(time) * 1000) / writeCount).toFixed(1)}ms each, ${versionStr})`
       );
     } else {
       ret.push(`(${versionStr})`);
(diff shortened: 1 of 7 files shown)


src/Eleventy.js
```diff
@@ -230,12 +230,16 @@ class Eleventy {
       `Wrote ${writeCount} ${simplePlural(writeCount, "file", "files")}`
     );
 
-    let time = ((new Date() - this.start) / 1000).toFixed(2);
+    let time = ((new Date().getTime() - this.start.getTime()) / 1000).toFixed(
```
zachleat: What’s the motivation for this change? I haven’t seen any bugs from this behavior

src/Eleventy.js
```diff
@@ -230,12 +230,16 @@ class Eleventy {
...
+      2
+    );
     ret.push(`in ${time} ${simplePlural(time, "second", "seconds")}`);
 
     if (writeCount >= 10) {
       ret.push(
-        `(${((time * 1000) / writeCount).toFixed(1)}ms each, v${pkg.version})`
+        `(${((Number(time) * 1000) / writeCount).toFixed(1)}ms each, v${
```
zachleat: Same as above, do we need the explicit cast here?

kleinfreund:
I’m referring to linter errors that my editor (VS Code) is showing me. Specifically errors shown by the built in TypeScript linter.

PR #724, opened 2019-10-09
Scope mapEntry variable to for..of loop in TemplateWriter

### LINKED ISSUE #722: TemplateWriterError reports the wrong template
**Describe the bug**
`TemplateWriterError`s thrown by `TemplateWriter#write` report the wrong template as the source of the error.

**To Reproduce**
In my case, the [eleventy-plugin-cache-buster](https://github.com/mightyplow/eleventy-plugin-cache-buster/) was throwing an error because an asset hadn't been written to the output folder before the plugin's transform function was run. 

However, [this loop](https://github.com/11ty/eleventy/blob/master/src/TemplateWriter.js#L176-L193) in `TemplateWriter.js` doesn't create a new scope to remember the value of the `mapEntry` variable with each loop iteration.

**Expected behavior**
The correct template should be reported. This could be solved by somehow creating a new scope with each loop iteration in the above mentioned loop.

**Environment:**
 - OS and Version: Mac
 - Eleventy Version 0.9.0

fredrikekelund:
Fixes #722. Previously we wouldn't scope the value of the mapEntry variable to each iteration when looping over the templates to write. This became an issue of any of the generated promises threw an error, since the error message would always report that it stemmed from the template from the last iteration.

### DIFF
--- a/src/TemplateWriter.js
+++ b/src/TemplateWriter.js
@@ -171,9 +171,8 @@ class TemplateWriter {
     await this._createTemplateMap(paths);
     debug("Template map created.");
 
-    let mapEntry;
     let usedTemplateContentTooEarlyMap = [];
-    for (mapEntry of this.templateMap.getMap()) {
+    for (let mapEntry of this.templateMap.getMap()) {
       promises.push(
         this._writeTemplate(mapEntry).catch(function(e) {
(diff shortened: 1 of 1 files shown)


zachleat:
Thank you!

Issue #820, opened 2019-12-12
Global data with 11tydata.json extension does not work


gloryofrobots:
**Describe the bug**

global data with 11tydata.json extension does not work due to the problem with TemplateData.getObjectPathForDataFile(); it returns ```filename.11tydata``` as a path for lodashset
instead of ```filename```

**To Reproduce**

create file test.11tydata.json in _data dir
with 
```
{
    "key":"value"
}
```
add  ```<h1> {{test.key}} </h1> ``` to your template
**Expected behavior**
```<h1>value</h1> ```must be rendered instead of ```<h1></h1>```

**Additional context**
I am working on PR for #794 and can fix this issue too. Or maybe this is correct behavior?
Please let me know

jevets:
This is the behavior I'd expect since it's `json` and not `js`. 

I'd name the file `test.json`. I've only ever used `11tydata.` extension prefix for `.js`.

Seeing `_data/test.11tydata.json`, I'd expect Eleventy to name that data `test.11tydata`, so I'd access `key` as `{{ test.11tydata.key }}` in templates.

For a JavaScript data file named `test.11tydata.js`, I'd access as `{{ test.key }}` in templates.

gloryofrobots:
```11tydata.js``` or  ```11tydata.json``` for global data it doesn`t matter. Global data object from 
 ```_data/test.11tydata.js``` will be accessible as **{{test.11tydata.key}}**  not **{{test.key}}** I just tested it.

jevets:
[Global data files](https://www.11ty.dev/docs/data-global/) don't expect the `.11tydata` extension prefixes. Eleventy doesn't remove the strings of the file name when naming the data object; it assumes you wanted the resulting data object to be named `test.11tydata` instead of `test`.

Examples:

- A `key` in a global data file named `_data/one.js` would be available as `{{ one.key }}` in templates.
- A `key` in a global data file named `_data/one.two.js` would be available as `{{ one.two.key }}` within templates.
- A `key` in a global data file named `_data/one/two.js` would also be available as `{{ one.two.key }}` from within templates. (may be a conflict there, not sure)
- A `key` in a global data file named `_data/one.two.three.four.js` would be available as `{{ one.two.three.four.key }}` within templates.

[...]

PR #821, opened 2019-12-12
Global data files priority according to file extensions  #794

### LINKED ISSUE #794: Global data doesn’t have explicit rules about data overrides when data file roots are the same
In template and directory data files override is explicit:

`component.11tydata.js` overrides `component.11tydata.json` overrides `component.json`

No such explicit convention exists for global data files. It does one generic glob selector for these files.

https://github.com/11ty/eleventy/blob/ccbaf5af8cafeccbf5b0ecde9e8b75c2219f07c3/src/TemplateData.js#L136

Surfaced by @gloryofrobots in #635.

gloryofrobots:
This PR resolves #794 and also adds merging of global data from different files with the same base filename into one object

For example if we have files
```
//_data/data.js
module.exports = {
  key1: "js-val1"
}

// _data/data.json
{
   "key1": "json-val1",
   "key2": "json-val2"
}
```
resulting data objects will be merged according to extension priority so ```js``` will override ```json```
```
{
  "key1":  "js-val1",
  "key2":  "json-val2"
}
```
Working on this PR I found that 11tydata.{json|js} files are not supported due to code in ```TemplateData.getObjectPathForDataFile```
I do not know if it is a bug or a normal behavior. To clarify this I created #820

### DIFF
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -2,6 +2,7 @@ const fs = require("fs-extra");
 const fastglob = require("fast-glob");
 const parsePath = require("parse-filepath");
 const lodashset = require("lodash/set");
+const lodashget = require("lodash/get");
 const lodashUniq = require("lodash/uniq");
 const merge = require("./Util/Merge");
 const TemplateRender = require("./TemplateRender");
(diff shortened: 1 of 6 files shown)


zachleat:
Please read https://github.com/11ty/eleventy/issues/820#issuecomment-568067392

zachleat:
As stated in #820 I have a few concerns that this will be a breaking change for folks? This might need to roll out in 1.0

gloryofrobots:
Thanks for the feedback.
This PR doesn't change how `TemplateData.getObjectPathForDataFile` works. So `_data/globalfile.11tydata.json`  will work as before.
Here I implemented the extension priority (so *js* will override *json*) and data merging according to this priority. Maybe, it will not break existing codebases because most of the projects, probably do not use the same filenames with different extensions for global data. In the current state, such data objects will override each other depending on the glob result order which is very unreliable in any case

zachleat:
> In the current state, such data objects will override each other depending on the glob result order which is very unreliable in any case.

Agree, yes. Makes good sense. I’ll remove the breaking change tag here.

zachleat:
Merging, thanks!

PR #960, opened 2020-02-23
Correct verb conjugation mistake in comment

samuelpath:
The past participle of the verb "override" is "overridden", not "overrode".
We say: "foo overrode bar", but "bar was overriden by foo".
https://www.theconjugator.com/english/verb/to+override.html


zachleat:
Thanks!

PR #967, opened 2020-02-26
Simplifying getTemplateData from 4 lines to 1

samuelpath:
This one-liner return is more concise and doesn't seem to make the reading more complex.
What do you think?

Another even more concise solution would be: 

`return this.templateData && new TemplateData(this.inputDir);`

But I'm not sure this would be as readable for readers not familiar with how "&&" works under the hood.

### DIFF
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -100,10 +100,7 @@ class EleventyFiles {
 
   // TODO make this a getter
   getTemplateData() {
-    if (!this.templateData) {
-      this.templateData = new TemplateData(this.inputDir);
-    }
-    return this.templateData;
+    return this.templateData ? this.templateData : new TemplateData(this.inputDir);
   }
 
   getDataDir() {


DirtyF:
https://twitter.com/codinghorror/status/1232103977178693632

samuelpath:
@DirtyF I didn't try to be fancy here, I genuinely find the one-line approach in my commit faster to reason about :)

src/EleventyFiles.js
```diff
@@ -100,10 +100,7 @@ class EleventyFiles {
 
   // TODO make this a getter
   getTemplateData() {
-    if (!this.templateData) {
-      this.templateData = new TemplateData(this.inputDir);
-    }
-    return this.templateData;
+    return this.templateData ? this.templateData : new TemplateData(this.inputDir);
```
pdehaan: I'm not sure this code does the same thing.
The old code would cache the value of `new TemplateData(this.inputDir)` into `this.templateData`, whereas I don't think this code does, so it's possible `this.templateData` would never get set, always returns a falsy value to the inline `if` and we'd be calculating `new TemplateData(this.inputDir)` each time.

samuelpath: Thanks @pdehaan, I missed the caching purpose of the code I was replacing!
Since the checks passed, I imagine that there's no test in place for this caching?

zachleat:
Appreciate these small pull requests @samuelpath! They are so much easier to review. Going to close this one though as I don’t think we want to change the code path there.

Just stylistically I usually find `if` to be easier to read than `ternary` code too if that helps. Although I’m personally not super strict about it.

Would be very interested in a PR that adds a test around this!

samuelpath:
Thanks @zachleat for your words of appreciation, you are a very generous maintainer :).

I will have a look at a PR to add a test around this path. It will allow me to move beyond minor stylistic PR, but this was a good entry point in this project for me. I'd like to be able to contribute more in the future.

PR #968, opened 2020-02-26
Inverse boolean logic to avoid empty statement

samuelpath:
I'm not sure what's the benefit of having a condition leading to an empty statement ?
You may have a good reason that I have missed though.


zachleat:
Thanks!

PR #969, opened 2020-02-26
Remove unused input param in compile function

samuelpath:
At first glance, we could imagine that keeping the unused parameter was a way to keep the compile function with a coherent signature across various engines, but some engines like Haml don't have the inputPath input parameter.


zachleat:
Ah yeah, this was an intentional leave—I want to keep it consistent!

PR #970, opened 2020-02-26
Make code style more coherent on error file

samuelpath:
In most files in the project, there is a line break between the const declarations on top and the class definition.

So I'm doing the same here to add some consistency.

The contrast is obvious when opening the 2 error files right after the other, as they should have the same layout but don't.


zachleat:
Thanks!

PR #987, opened 2020-03-02
Add inputPath to Haml eng. to improve consistency

samuelpath:
In pull request #969, you said:

> Ah yeah, this was an intentional leave—I want to keep it consistent!

So you decided to leave the unused `inputPath` param in the Handlebars engine in order to be consistent.

So following this logic, it would make sense to **add** this same unused `inputPath` param in the Haml engine's compile function, right?

As you can see from the search below, the Haml engine's compile function is the only one missing this input param:

image

Right?


zachleat:
Thank you!

PR #992, opened 2020-03-03
Replace concat by unshift as asked in comment

samuelpath:
Your comment (that made me smile by the way) said: 

> // todo use unshift or something (no wifi here to look up docs :D)

So here we go :).

PS. If you want to have the JS docs available offline, have a look at https://devdocs.io/offline.

### DIFF
--- a/src/TemplateRender.js
+++ b/src/TemplateRender.js
@@ -112,8 +112,7 @@ class TemplateRender {
 
     // markdown should always be first
     if (usingMarkdown) {
-      // todo use unshift or something (no wifi here to look up docs :D)
-      engines = ["md"].concat(engines);
+      engines.unshift("md");
     }
 
     return engines;


zachleat:
Thanks! Forgot about that 😅

samuelpath:
@pdehaan Thanks for this tip. If @zachleat thinks this is a good idea, you could submit a PR to make running ESLint easier (or even automatic/by default?), and add this rule, what do you think?

It depends how much of a "technical debts" these "TODO"s feel or truly are though. For instance, replacing a concat by unshift is nice but is not a big deal compared to shipping new features or fixing real bugs.

samuelpath:
@pdehaan Actually, we already know that @zachleat likes ESLint, thanks to the following [commit](https://github.com/11ty/eleventy/commit/a0c7c1899f315bd9b7530118610ee5d7e859b981):

image

So you can safely assume that any PR to make the use of ESLint for this project smoother should be warmly welcomed :).

PR #998, opened 2020-03-05
Add a cache for getTemplates as asked in TODO

samuelpath:
In the `getTemplates` function, @zachleat had a comment saying `// TODO cache this`.

I've implemented my understanding of how the cache should work, if this can be helpful in any way.

If my PR gets accepted, I will add a unit test that tests this caching logic.

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -425,9 +425,16 @@ class Template extends TemplateContent {
       );
     }
   }
+  
+  setTemplates(templates) {
+    this.templates = templates;
+  }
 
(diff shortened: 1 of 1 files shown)


samuelpath:
Oups, I've added a caching mechanism without taking into account the function input param. I'll quickly fix this!

zachleat:
Ah—sorry I believe this was implemented in already merged https://github.com/11ty/eleventy/pull/1273

I will remove that comment though!

PR #999, opened 2020-03-05
Use getter/setter for getTemplateData

samuelpath:
In a comment above the `getTemplateData` function, @zachleat said:

> // TODO make this a getter

So please find here my humble attempt to solve this TODO.


zachleat:
Thanks!

PR #1006, opened 2020-03-06
Refactor cmd.js to use minimist options.unknown callback ("TODO" fix)

samuelpath:
In cmd.js, there was a comment saying:

> TODO refactor to use minimist options.unknown callback

The [minimist doc](https://github.com/substack/minimist) says:

> opts.unknown - a function which is invoked with a command line parameter not defined in the opts configuration object. If the function returns false, the unknown option is not added to argv.

This means that if we use this feature from minimist, the `EleventyCommandCheck` class with all its helper functions becomes obsolete (this removes 75 lines of code). This also means that the test file `EleventyCommandCheckTest.js` becomes also obsolete, which means one less test file to run.

Since the `EleventyCommandCheckError` class was in the `EleventyCommandCheck.js` file, I renamed that file `EleventyCommandCheckError.js` while only keeping the bare minimum inside.

I have tested manually and it runs as expected (there is no unit test file for `cmd.js` since this is not a module I guess).


zachleat:
Thank you!

PR #1007, opened 2020-03-06
Move Template.write to TemplateTest, since only used there ("TODO" fix)

samuelpath:
In `Template.js`, @zachleat asked of the `write` function:

> // TODO is this still used by anything but tests?

The answer is clearly no, so this function belongs to `TemplateTest.js` instead (the same way the local helper function `getRenderedData` does).

So I cut the function from `Template.js` where it was unused and pasted it to `TemplateTest.js`.

I had some adaptations to make regarding the input parameters and the value of `this`, as is often the case when porting a function from a file to another.

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -584,17 +584,6 @@ class Template extends TemplateContent {
     return Promise.all(promises);
   }
 
-  // TODO is this still used by anything but tests?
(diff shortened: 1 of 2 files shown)


zachleat:
Thank you!

PR #1010, opened 2020-03-09
Improve template clone by using constructor logic ("TODO" fix)

samuelpath:
Above the `clone()` function in `Template.js`, @zachleat said:

> // TODO this but better

I wasn't exactly sure what was meant by "better", so I tried to implement a way that seems slightly cleaner to me. But since cleanliness in code is subjective, I might have missed what Zach truly meant by "better" :).

Now the `clone()` function simply returns a new Template, using the template to be cloned's values as input parameters. I've then updated the constructor to take into account these new params, while also defaulting them to the value initially set in the constructor before some of them where parameterized.

Before doing this refactor, I tried to go even further by ditching the `clone()` function altogether, and replace all its calls by Lodash's `cloneDeep`, but it messed up plenty of tests by throwing a `TemplateContentPrematureUseError` error. So it seems to me that we don't want a perfect deep clone, but I still haven't figured out why, and in what ways our clone passes the tests but not Lodash's…

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -19,7 +19,17 @@ const debug = require("debug")("Eleventy:Template");
 const debugDev = require("debug")("Dev:Eleventy:Template");
 
 class Template extends TemplateContent {
-  constructor(path, inputDir, outputDir, templateData) {
+  constructor(
+    path,
+    inputDir,
+    outputDir,
+    templateData,
+    config,
+    transforms = [],
+    linters = [],
+    isVerbose = true,
+    isDryRun = false
+  ) {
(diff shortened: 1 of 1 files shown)


samuelpath:
I'm not quite sure why all the tests pass except on Mac with Node 12…

It's the `ResetBenchmark` test from `BenchmarkTest.js` that returns 2 instead of 0.

If anyone has a clue…

zachleat:
Hey I’m going to close this one because I’m not a fan of huge constructor signatures. Still looking for a better way to do this. Appreciate all of the PRs though! 🏆

Issue #1012, opened 2020-03-09
Validate Javascript Templates


ithinkihaveacat:
If a Javascript Template file exists (e.g. `foo.11ty.js`), and it is specified (via `layout: foo`) but does not actually return anything useful (for example, if it's an empty file), eleventy returns a cryptic `fn is not a function` error.

It would be helpful if instead eleventy checked to see if the template file returned a valid value, such as a raw value, function, or class.

zachleat:
Closed by #1013

PR #1013, opened 2020-03-09
Throw error if nothing useful is returned from JS template

ithinkihaveacat:
Fixes `fn is not a function` error (#1012).


zachleat:
Thank you!

Issue #469, opened 2019-03-29
Upgrade Liquid major version


zachleat:
We are on v6, they are on v8.

https://github.com/harttle/liquidjs/blob/master/CHANGELOG.md

Breaking changes apply to options name changes (no template facing changes here).

See also https://github.com/harttle/liquidjs/issues/109

paulrobertlloyd:
Any reason for this not being in the v0.8.0 beta? I mis-read the release notes, and thought it had bumped Liquid to the latest major version, but was actually a minor one. I only ask as [I came across an issue](https://github.com/harttle/liquidjs/issues/117), and having certainty about the version used by 11ty (having tried to install the latest major version separately) would help me resolve it.

zachleat:
The reason is that it would break compatibility if anyone is using the options keys that were renamed. See https://www.11ty.io/docs/languages/liquid/#optional%3A-use-your-own-options

zachleat:
Hmm It’s possible we could shim those—I’ll have to think about it.

kleinfreund:
@zachleat I think this should just be scheduled with the next major version without shimming it for now.

#421 is also a breaking change and the follow-up PR to that will be, too. So releasing a major release could be valuable soonish?

paulrobertlloyd:
@zachleat Investigating using my own instance of Liquid, thought it was worth recording the raw build times seen when swapping out the built in version for the latest release.

With builtin LiquidJS instance (6.4.3), 5 runs:
* `Copied 4 items / Wrote 1463 files in 15.73 seconds (10.8ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 16.11 seconds (11.0ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 15.66 seconds (10.7ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 16.02 seconds (11.0ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 15.80 seconds (10.8ms each, v0.10.0)`

Average build time: **15.86 seconds**

With own LiquidJS instance (9.6.2), 5 runs:
* `Copied 4 items / Wrote 1463 files in 11.82 seconds (8.1ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 12.06 seconds (8.2ms each, v0.10.0)`
* `Copied 4 items / Wrote 1463 files in 11.72 seconds (8.0ms each, v0.10.0)`
[...]

Issue #212, opened 2018-09-08
Support for macro imports in parent templates


edwardhorsford:
I'm coming from a node / nunjucks app where I import my macros once in my parent layout template like so:

`{% import "includes/macros.html" as macro with context %}`

Now all pages that use this template get access to my macros.

I couldn't get this to work with eleventy - it *appears* like every page needs to reference the macros file individually. It would be great if they could be referenced just the once.

Some conversation on twitter about this:
https://twitter.com/adambsilver/status/1038330552363966464

zachleat:
Is your node/nunjucks app using Nunjucks `extends` for layouts? We do still support those.

Hmm, this one actually might be difficult to get working with eleventy layouts. Structurally, we render layout-less output of templates for `templateContent` strings in collection entries.
https://www.11ty.io/docs/collections/#collection-item-data-structure

edwardhorsford:
No, I'm using the Eleventyone project as a starter. I can give extends a try - that's what my existing site uses.

Is the an advantage / disadvantage to using one or the other?

zachleat:
If you use Nunjucks extends for layouts, you can’t use front matter in your layout to set inherited data. But you can use directory data files to do this same thing https://www.11ty.io/docs/data-template-dir/

The other thing Eleventy layouts gives you is cross-template layout use. Eleventy layouts can use any template language, independent of the child content. This may or may not be important for your project.

edwardhorsford:
I've tried using extends in my layouts and I *think* it still doesn't work when importing macros in the parent template.

zachleat:
Sounds like this is going to need a new feature to support this request. Possibly a front matter convention—not sure yet? I’m open to ideas!

edwardhorsford:
I've ended up including the import on every page, including templates. It's not a major thing, but a bit annoying.

Thankfully eleventy errors obviously well when it's *not* included, so I know pretty quickly when I've messed it up.

Of course my ideal would be that each page only has what's *unique* about that page - with imports and such coming from higher level templates. But I'm not sure it's worth adding more magic to frontmatter to achieve this.

paulmsmith:
+1 To this request.

I've tried a few tricks to get this working but was fighting Eleventy layouts. I tried creating a nunjucks filter that exposed the `addGlobal` method but it didn't work, probably with how eleventy preps the templates for render.

mashdot:
[...]

PR #1058, opened 2020-03-29
chore: upgrade 2x performant LiquidJS v9

harttle:
There're substantial performance improvements in v9 and many bugfixes from v6 to v9. There're some API changes and some of them can be indicated in this PR:

* inside custom tag implementation, `engine.renderer.renderTemplates()` can be used only in generator tags as you can see in `addPairedShortcode()`. The difference is subtle here but generator tags provides [a major performance boost](https://github.com/harttle/liquidjs/pull/166) and provide code reuse between sync/async.
* `evalValue` is now changed to async and another `evalValueSync()` is added for sync calls.
* some methods (see the following`evalValue()` `in parseArguments()`) are nolonger static.

A good thing I find is that for people using Liquid via `setEngineLib()` will not be affected. 

Another thing is, the patch is automatically formatted by your git hooks, is it supposed to be this way or should I disable those hooks and re-submit? Thank you for you time!


zachleat:
Beautiful, thank you!

Are you saying this isn’t a breaking change for any of the template language syntax?

zachleat:
Fixes #469

harttle:
> Are you saying this isn’t a breaking change for any of the template language syntax?

There should be no breaking changes except for some undocumented features (I mean some fault tolerance provided by LiquidJS but not by Shopify/liquid or Jekyll like [#212](https://github.com/harttle/liquidjs/issues/212)). Additionally, the filters should be more robust after all these bug fixes.

aciccarello:
Updating Liquid would also update strftime used by the date filter. I was confused why the strftime docs didn't match what I was seeing.
harttle/liquidjs#77

harttle:
@aciccarello Thank you to point it out. I've updated the doc for [more specific description of strftime](https://github.com/harttle/liquidjs/commit/2fa4edaa84ea1cd76c27a171f0719105f94878f9). Basically it should be conformant to Ruby's core [DateTime#strftime](https://apidock.com/ruby/DateTime/strftime). If there's still anything unexpected, feel free to open an issue on [harttle/liquidjs](https://github.com/harttle/liquidjs/issues).

zachleat:
Shipping this! Thank you!

Issue #495, opened 2019-04-15
Feature request: Expose Nunjucks' addGlobal


robdodson:
I was curious if [addGlobal](https://github.com/mozilla/nunjucks/pull/239) might be something eleventy could expose.

In my instance I wanted to call a function that takes two arguments. It doesn't feel like I could do this correctly with a filter. I could use a shortcode to do this, but then I can't pipe the results into a filter.

Essentially I'm trying to do:

```
{{ findItemInCollection(collection, slug) | dump }}
```

Ryuno-Ki:
Would a syntax like `{{ collection | findItem(slug) | dump }}` work for you?

(I believe I implemented something related to that but haven't packaged it up, yet)

robdodson:
I think that might work but it's sort of arcane syntax. Kind of related to [this comment from the nunjucks maintainer](https://github.com/mozilla/nunjucks/issues/247#issuecomment-96900678) that filters are somewhat awkward syntax on top of function calls.

edwardhorsford:
@robdodson Elventy might not directly support it, but you can add them manually (I discovered this week) by using your own Nunjucks environment. I'm already doing similar to be able to set some specific Nunjucks options.

## Inside your eleventy.js:
```js
  let nunjucksEnvironment = new Nunjucks.Environment(
    new Nunjucks.FileSystemLoader(["src/site/_templates"]), // we need to pass in our includes dir here
    { 
      lstripBlocks: true,
      trimBlocks: true,
      autoescape: false
     });

  nunjucksEnvironment.addGlobal(name, nunjucksGlobals[name]);

  eleventyConfig.setLibrary("njk", nunjucksEnvironment);
```

FWIW I've made a similar filter. I'm pleased to see the syntax matches @Ryuno-Ki's suggested. I also accept an optional arg to do exact matching or not.

I've also just found out that filters (at least for Nunjucks) have access to the context. So I'm tempted to rework my filter to not require collections at all - just take a slug and return the item / array of items / false.

zachleat:
> This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

> View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

zachleat:
[...]

PR #1060, opened 2020-03-29
allow registering Nunjucks globals

lxg:
This PR allows registering Nunjucks globals through the eleventy config. The purpose of this is mainly to allow creating global template functions where filters and tags aren’t enough. 

A simple example:

in .eleventy.js
```js
eleventyConfig.addNunjucksGlobal("foobar", (str) => `foo ${str} bar`)
```

usage in a template file:
```nunjucks
{{ foobar("hello world") }}
```

Result:
```nunjucks
foo hello world bar
```

### DIFF
--- a/src/Engines/Nunjucks.js
+++ b/src/Engines/Nunjucks.js
@@ -36,6 +36,7 @@ class Nunjucks extends TemplateEngine {
       this.config.nunjucksAsyncPairedShortcodes,
       true
     );
+    this.addGlobals(this.config.nunjucksGlobals);
   }
 
   addFilters(helpers, isAsync) {
@@ -61,6 +62,16 @@ class Nunjucks extends TemplateEngine {
     }
 
     this.njkEnv.addExtension(name, tagObj);
+  }
+
+  addGlobals(globals) {
+    for (let name in globals) {
+      this.addGlobal(name, globals[name]);
+    }
+  }
+
+  addGlobal(name, globalFn) {
+    this.njkEnv.addGlobal(name, globalFn);
   }
 
   addAllShortcodes(shortcodes, isAsync = false) {
--- a/src/UserConfig.js
+++ b/src/UserConfig.js
@@ -29,6 +29,7 @@ class UserConfig {
     this.nunjucksFilters = {};
     this.nunjucksAsyncFilters = {};
     this.nunjucksTags = {};
+    this.nunjucksGlobals = {};
     this.nunjucksShortcodes = {};
     this.nunjucksAsyncShortcodes = {};
     this.nunjucksPairedShortcodes = {};
@@ -224,6 +225,21 @@ class UserConfig {
     }
 
     this.nunjucksTags[name] = bench.add(`"${name}" Nunjucks Custom Tag`, tagFn);
+  }
+
+  addNunjucksGlobal(name, globalFn) {
+    name = this.getNamespacedName(name);
+
+    if (this.nunjucksGlobals[name]) {
+      debug(
+        chalk.yellow(
+          "Warning, overwriting a Nunjucks global with `addNunjucksGlobal(%o)`"
+        ),
+        name
+      );
+    }
+
+    this.nunjucksGlobals[name] = bench.add(`"${name}" Nunjucks Global`, globalFn);
   }
 
   addTransform(name, callback) {
@@ -649,6 +665,7 @@ class UserConfig {
       nunjucksFilters: this.nunjucksFilters,
       nunjucksAsyncFilters: this.nunjucksAsyncFilters,
       nunjucksTags: this.nunjucksTags,
+      nunjucksGlobals: this.nunjucksGlobals,
       nunjucksAsyncShortcodes: this.nunjucksAsyncShortcodes,
       nunjucksShortcodes: this.nunjucksShortcodes,
       nunjucksAsyncPairedShortcodes: this.nunjucksAsyncPairedShortcodes,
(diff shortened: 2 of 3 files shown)


lxg:
btw, fixes #495

nhoizey:
@lxg are these like Nunjucks macros?

lxg:
@nhoizey Not exactly. While they both behave like funktions, Nunjucks macros are defined with Nunjucks syntax, whereas you would write such a *global* as Javascript and just expose the functionality to Nunjucks.

For example:

```nunjucks
{# this is nunjucks syntax #}
{% macro hello(who) %}
    Hello {{ who }}
{% endmacro %}
```
vs.

```js
eleventyConfig.addNunjucksGlobal("hello", who => `Hello ${who}`)
```

Both would expose the same functionality in this simple example, but obviously the JS native is more powerful in regard to syntax and context.

Also, and this is my primary reason for this PR, it gives you more power than filters and tags in Nunjucks. In my scenario, there are translation functions that take parameters and are wrapped in `sprintf`s … something you can hardly express with filters.

nhoizey:
Ok, I understand the difference now, thanks!

It looks like it could fix my issue with tags not getting "external" data, leading to the need to pass information from one to another.

For example, I use this macro in my article template:
https://github.com/nhoizey/nicolas-hoizey.com/blob/master/src/_includes/macros/articleCard.njk

Which calls this other macro, which is called also for other content types:
https://github.com/nhoizey/nicolas-hoizey.com/blob/master/src/_includes/macros/meta.njk

I have to pass data and set default values for different use cases, that's not really easy and clean.

Globals look better.

Dexus:
ping

lxg:
@zachleat any opinions regarding this PR?

spl:
I was just looking for `addGlobal`, too. `addNunjucksGlobal` seems like a useful and straightforward addition. But if this PR isn't accepted, I think the problem can also be solved by adapting the [Use Your Nunjucks Environment instructions](https://www.11ty.dev/docs/languages/nunjucks/#optional-use-your-nunjucks-environment) with this `.eleventy.js`:

```js
let Nunjucks = require('nunjucks')

module.exports = (config) => {
  let nunjucksEnvironment = new Nunjucks.Environment()
  nunjucksEnvironment.addGlobal('foobar', (str) => `foo ${str} bar`)
  config.setLibrary('njk', nunjucksEnvironment)
}
```

edwardhorsford:
I do what @spl suggested - making my own environment and passing globals to it.

zachleat:
Shipping this with 1.0, thank you!

gremo:
Any chance to use it with the current version of 11ty? Thanks...

Ryuno-Ki:
@gremo You could check out via git branch (but that'll pull in other changes, too).

pdehaan:
Q: Does `eleventyConfig.addNunjucksGlobal()` support passing direct values (ie: `eleventyConfig.addNunjucksGlobal("foobar", "hello!")`) or do I have to wrap it in a function?

This seems to work in Nunjucks (see [RunKit](https://runkit.com/embed/5v6ew7zdox1q)):
```js
const nunjucks = require("nunjucks");
const env = new nunjucks.Environment();
env.addGlobal("cats", 42);
const res = env.renderString('{{ username }} owns {{ cats }} cats!', { username: 'James' });
console.log(res); // Output: "James owns 42 cats!"
```

But I get errors in Eleventy:

```js
// .eleventy.js snippet
eleventyConfig.addNunjucksGlobal("cats", 42);
```

```njk
{{ cats }}
```

### OUTPUT
```
function (...args) {
      benchmark.before();
      let ret = callback.call(this, ...args);
      benchmark.after();
      return ret;
    }
```

But then if I try:

```njk
{{ cats() }}
```

I get a build error with the following:

```
Problem writing Eleventy templates: (more in DEBUG output)
> Having trouble rendering njk template ./src/pages/home.njk

`TemplateContentRenderError` was thrown
> (./src/pages/home.njk) [Line 4, Column 7]
  TypeError: callback.call is not a function

`Template render error` was thrown:
    Template render error: (./src/pages/home.njk) [Line 4, Column 7]
      TypeError: callback.call is not a function
        at Object._prettifyError (/private/tmp/11ty-nunjucks-globals/node_modules/nunjucks/src/lib.js:36:11)
        at /private/tmp/11ty-nunjucks-globals/node_modules/nunjucks/src/environment.js:563:19
        at Template.root [as rootRenderFunc] (eval at _compile (/private/tmp/11ty-nunjucks-globals/node_modules/nunjucks/src/environment.js:633:18), <anonymous>:42:3)
        at Template.render (/private/tmp/11ty-nunjucks-globals/node_modules/nunjucks/src/environment.js:552:10)
        at /private/tmp/11ty-nunjucks-globals/node_modules/@11ty/eleventy/src/Engines/Nunjucks.js:451:14
        at new Promise (<anonymous>)
        at /private/tmp/11ty-nunjucks-globals/node_modules/@11ty/eleventy/src/Engines/Nunjucks.js:450:14
        at Template._render (/private/tmp/11ty-nunjucks-globals/node_modules/@11ty/eleventy/src/TemplateContent.js:310:28)
        at async Template.getTemplateMapContent (/private/tmp/11ty-nunjucks-globals/node_modules/@11ty/eleventy/src/Template.js:953:19)
        at async TemplateMap.populateContentDataInMap (/private/tmp/11ty-nunjucks-globals/node_modules/@11ty/eleventy/src/TemplateMap.js:459:39)

Wrote 0 files in 0.07 seconds (v1.0.0-canary.39)
```

Workaround seems to be to always wrap the global value in a function, which might be inconsistent w/ Nunjucks behavior:

```js
// .eleventy.js snippet
eleventyConfig.addNunjucksGlobal("cats", () => 42);
```

Then use `{{ cats() }}` which will output `42` as expected.

zachleat:
I pushed a fix and tests for the issue with using a literal function @pdehaan  https://github.com/11ty/eleventy/pull/1060#issuecomment-881936530

PR #1065, opened 2020-04-04
Fix "[object Object]" in TemplatePassthrough debug message

stevenpetryk:
No big thang, just something I noticed when running in debug mode. It's useful to see this path properly.

### After

image


zachleat:
Thanks!

PR #1066, opened 2020-04-04
Prevent build errors from breaking the watcher

### LINKED ISSUE #833: Syntax errors in global data files breaks the watcher
**Describe the bug**

I am using a global data file inside a `_data` folder to share global links and info on the website.

When running eleventy in watch mode, if when I save the file the content has a syntax error the watcher errors and seems to stop.

While this is totally fine in a build process I am not sure it's so good during development.  

**To Reproduce**

1. Add a new global data file `demo.js` with the content:

```js
module.exports = {
    hello: 'worldx'
}
```

2. Print the hello property data in a Nunjucks template: `{{ demo.hello }}`
3. Change the content to something broken:

```js
module.exports = {
    hello 'worldx // removed ending '
}
```

4. Console errors:

```
Eleventy fatal watch error: (more in DEBUG output)
> Invalid or unexpected token (SyntaxError):
        hello: 'world
               ^^^^^^

    SyntaxError: Invalid or unexpected token
```

5. Fixing the error does not trigger the watcher anymore

**Expected behavior**

I'd expect the watcher to be still running

**Screenshots**

**Environment:**

- OS and Version: Mac 10.14
- Eleventy Version 0.8.3

**Additional context**

stevenpetryk:
I think this will close #833, but there are a few concerns I have:

1. I'm not sure if this is the correct way to "abort" this particular run.

    Should I be wrapping this whole call in a try/catch? Should I specifically catch the error that's raised instead? My thinking is that no build error should crash the watcher.

1. I have no idea how to test this. Guidance is welcome!

## In action

Introducing a syntax error shows the exception, aborts the build, and doesn't livereload, but it also keeps the watcher running. Fixing the error then continues the build as normal.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -451,7 +451,13 @@ Arguments:
       this.resetConfig();
     }
 
-    await this.restart();
+    try {
+      await this.restart();
+    } catch (e) {
+      EleventyErrorHandler.error(e, "Eleventy watch error");
+      this.watchManager.setBuildFinished();
+      return;
+    }
 
     this.watchTargets.clearDependencyRequireCache();
 


stevenpetryk:
Hmm, not entirely sure this is a good approach after digging into the code more and writing the error overlay plugin.

stevenpetryk:
Replaced by https://github.com/11ty/eleventy/pull/1071

Issue #359, opened 2018-12-31
Example: use GitHub actions to deploy Eleventy to GitHub pages


zachleat:
reported by https://twitter.com/acwervo/status/1079502388707250178

zachleat:
This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for documentation requests. This means documentation issues will now be closed instead of leaving them open.

View [the documentation queue backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=is%3Aissue+sort%3Areactions-%2B1-desc+is%3Aclosed+label%3Adocumentation+label%3Aneeds-votes). Don’t forget to upvote the top comment with 👍!

TartanLlama:
I've made a GitHub Action which runs Eleventy and has an example of how to also automatically publish to GitHub Pages on each push: https://github.com/TartanLlama/actions-eleventy/

jonasjohansson:
I get build error due to unrecognized tags (I have custom 11ty tags and not Liquid), any clue what to do?

christian-fei:
This could help @jonasjohansson  @TartanLlama https://christianfei.com/posts/2019-08-29-Deploy-Eleventy-site-with-Github-Actions-on-AWS-S3/

fogrew:
@TartanLlama I wanted to, but could not take advantage of your action. It would be cool [to use a new approach to API keys](https://github.com/TartanLlama/actions-eleventy/issues/1).

As long as the option above did not work for me, I had to use [my own](https://github.com/fogrew/gurylev.com/blob/master/.github/workflows/eleventy-build.yml). It looks simple enough.

zachleat:
https://www.11ty.dev/docs/deployment/#mini-tutorials

PR #1067, opened 2020-04-04
Fix roadmap links in the README

stevenpetryk:
The "needs-votes" label was renamed to "enhancement: needs-votes", which has broken these links in the README and has also broken links in a bunch of @zachleat's "lodash-style issue management" comments ([example](https://github.com/11ty/eleventy/issues/359#issuecomment-452366352)).

I can at least fix the first thing, I think Zach will have to change the comments and the saved reply.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,10 @@ npm run test
 
 ## Major Roadmapped Features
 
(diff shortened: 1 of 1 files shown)


zachleat:
oh well, shit. Thanks for reporting. I think it’s probably not worth the rename, I’ll reinstate the old label name for now.

PR #1070, opened 2020-04-05
Add error overlay to the core

### LINKED ISSUE #525: Error-out on failure with `eleventy --serve`
Since `eleventy --serve` continues to attempt to generate output and does not error-out on failures, I'm often confused by how making changes doesn't seem to reflect on the output only to realise that an error was encountered long back but the server continued to serve an old copy of the output. I use [hotel](//github.com/typicode/hotel), which will display a clear error message in the browser when attempting to access the site after the server crashes with an non-zero error code. 

Having `eleventy --serve` error-out on failure (at least by default) would be awesome.

stevenpetryk:
Closes #525. Bringing [eleventy-plugin-error-overlay](https://github.com/stevenpetryk/eleventy-plugin-error-overlay) into the core.

The main way I went about implementing this was with the EventEmitter pattern, which is already present in some parts of the code.

- `beginWatch` on the EleventyConfig events clears the error object
- A new event, `errorLogged`, on the ErrorHandler, is used to set the object (this way we have full parity between what's logged and what goes to the browser)

The middleware does something a bit hacky: it's initialized with a getter callback that is expected to return an error or null. This is so that EleventyServe can easily forward error information to it without having to pass something mutable or pass itself. Happy to rework this if it seems warranted.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -548,6 +548,14 @@ Arguments:
     benchmark.before();
     await this._initWatchDependencies();
     benchmark.after();
+
+    // Error overlay in development
+    this.config.events.on("beforeWatch", () => {
+      this.eleventyServe.clearError();
+    });
+    EleventyErrorHandler.events.on("errorLogged", error => {
+      this.eleventyServe.showError(error);
+    });
   }
 
   /**
--- a/src/Middleware/ErrorOverlayMiddleware.js
+++ b/src/Middleware/ErrorOverlayMiddleware.js
@@ -1,0 +1,145 @@
+const ErrorStackParser = require("error-stack-parser");
+const debug = require("debug")("Eleventy:ErrorOverlayMiddleware");
+
+/**
+ * A browser-sync middleware that pipes build errors to a browser view regardless of the requested
+ * path.
+ *
+ * @param {() => Error | null} getError a callback evaluated on every request that should return an
+ *  error to be shown (if any). Returning `null` bypasses the middleware.
+ *
+ * @returns {Function} the middleware
+ */
+module.exports = function ErrorOverlayMiddleware(getError) {
+  return function(_req, res, next) {
+    const error = getError();
+
+    if (!error) {
+      next();
+      return;
+    }
+
+    debug("Showing error", error);
+    const stacktrace = ErrorStackParser.parse(error).map(item => ({
+      ...item,
+      fileName: item.fileName
+        .replace(process.cwd(), ".")
+        .replace(process.env.HOME, "~")
+    }));
+
+    res.status = 500;
+    res.setHeader("content-type", "text/html");
+    res.end(
+      template({
+        errorName: error.name,
+        errorMessage: error.message,
+        stacktrace
+      })
+    );
+  };
+};
+
+const template = ({ errorName, errorMessage, stacktrace }) => `
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>${errorName}: ${errorMessage}</title>
+
+    <style>
+      :root {
+        --foreground: 0, 0, 0;
+        --background: 255, 255, 255;
+      }
+
+      @media (prefers-color-scheme: dark) {
+        :root {
+          --foreground: 204, 204, 204;
+          --background: 34, 34, 34;
+        }
+      }
+
+      *,
+      *:before,
+      *:after {
+        box-sizing: border-box;
+        position: relative;
+        margin: 0;
+        padding: 0;
+      }
+
+      body {
+        border-top: 0.5em solid rgb(197, 26, 26);
+        font-family: system-ui;
+        font-size: 18px;
+        background: rgb(var(--background));
+        color: rgb(var(--foreground));
+      }
+
+      main {
+        padding: 4em 2em;
+        max-width: 40em;
+        margin: 0 auto;
+      }
+
+      main > * + * {
+        margin-block-start: 2em;
+      }
+
+      code {
+        font-size: inherit;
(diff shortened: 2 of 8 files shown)


src/Eleventy.js
```diff
@@ -548,6 +548,14 @@ Arguments:
     benchmark.before();
     await this._initWatchDependencies();
     benchmark.after();
+
+    // Error overlay in development
+    this.config.events.on("beforeWatch", () => {
+      this.eleventyServe.clearError();
+    });
+    EleventyErrorHandler.events.on("errorLogged", error => {
```
stevenpetryk, line 556: Why not just name the error `error`? It turns out that `EventEmitter` will automatically throw errors attached to events named `error`. Lots of confusion resulted from that behavior :)

src/Middleware/ErrorOverlayMiddleware.js
```diff
@@ -0,0 +1,145 @@
...
+      template({
+        errorName: error.name,
+        errorMessage: error.message,
+        stacktrace
+      })
+    );
+  };
+};
+
+const template = ({ errorName, errorMessage, stacktrace }) => `
```
stevenpetryk, line 42: I chose to implement this as a JS template to avoid pulling in `nunjucks` just to render one small template. I also didn't know where I would even put a nunjucks file in this hierarchy. Feedback welcome here :)

test/EleventyServeTest.js
```diff
@@ -29,6 +30,7 @@ test("Get Options", t => {
     server: {
       baseDir: "_site"
     },
+    middleware: options.middleware,
```
stevenpetryk, line 33: This is a little jank. ava's `deepEqual` compares functions by identity, and there's no way to get two copies of the same function in this scenario since the middleware always returns a new function. So, I'm left with "opting out" of the deep compare here.

Ryuno-Ki: Hm … you could cast a function to a string (via `+ ''`) and compare the function body :-D

stevenpetryk: Yeah but the “actual” object will still have a function there. I don’t think I can make ava cast that, right?

Ryuno-Ki: I wouldn't know how :-/

stevenpetryk:
Are we supposed to request a review or anything to get PRs merged? Just want to make sure since it's been a minute :)

aharvard:
Excited to see this submitted to core! 

I was just setting up the plugin into my site and wanted to make sure we have a test case covered. Or, maybe I misconfigured and this is not a real issue, if so plz disregard. 

After install and config setup, I intentionally introduce an problem that threw this error:
```
File changed: src/06_components/form-offset/demo-default.njk
Problem writing Eleventy templates: (more in DEBUG output)
> Having trouble rendering njk template ./src/06_components/form-offset/demo-default.njk

`TemplateContentRenderError` was thrown
> (./src/06_components/form-offset/demo-default.njk)
  Error: template not found: /Users/c3hy0/sites/rdc-ui/src/06_components/form-offset/_form-partial

`Template render error` was thrown:
    Template render error: (./src/06_components/form-offset/demo-default.njk)
      Error: template not found: /Users/c3hy0/sites/rdc-ui/src/06_components/form-offset/_form-partial
        at Object._prettifyError (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/lib.js:36:11)
        at /Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:561:19
        at eval (eval at _compile (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:631:18), <anonymous>:41:11)
        at createTemplate (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:293:11)
        at next (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/lib.js:282:7)
        at handle (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:327:11)
        at /Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:337:9
        at next (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/lib.js:280:7)
        at Object.asyncIter (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/lib.js:286:3)
        at Environment.getTemplate (/Users/c3hy0/sites/rdc-ui/node_modules/nunjucks/src/environment.js:319:9)
Copied 1 item / Wrote 0 files in 2.36 seconds (v0.10.0)
```
For me, the critical info is this line: 
```
Error: template not found: /Users/c3hy0/sites/rdc-ui/src/06_components/form-offset/_form-partial
```
However in the browser I did not see it: 
image

is the idea that `errorLogged` will ensure we get all the content in browser? 
> A new event, errorLogged, on the ErrorHandler, is used to set the object (this way we have full parity between what's logged and what goes to the browser)

Thanks!

stevenpetryk:
I'll bet that a different exception is thrown at the watcher level than at the rendering level, so what you're probably seeing in the overlay is the "outermost" error, but in the console you're seeing the real error.

I'm not sure if this discrepancy is worth fixing in this PR 🤔Zach might have an opinion though.

zachleat:
Sorry for the late response here—I’ve resolved the conflicts at https://github.com/11ty/eleventy/tree/stevenpetryk-sgp/error-overlay if you want to update.

Again I think this is really great but needs some minor tweaks?

Specifically the console shows better contextual information than the overlay:

image

versus

image

zachleat:
Note that I’m open to changing how error handling works if there are improvements to be made there

Ryuno-Ki:
I'm inclined to skip internal / non-Eleventy lines from the output (the GUI).
Would it be possible to catch the signature (or even passed in arguments?)

PR #1071, opened 2020-04-06
Prevent unhandled errors from breaking watch mode

### LINKED ISSUE #833: Syntax errors in global data files breaks the watcher

stevenpetryk:
This closes #833 and replaces #1066 as a solution. When I did some digging, I found out that the real problem here is that some errors are never caught by the `watchRun` callback. The `try` / `catch` is unused because the error is thrown inside of a `setTimeout`.

The fact that an exception is thrown means the build never gets marked as finished, causing `_watch` to do nothing:

https://github.com/11ty/eleventy/blob/8ffbb9927ea8482c5a46515f3a2ed25cd8830ad8/src/Eleventy.js#L441-L443

_But_ since the `try` / `catch` also doesn't work, Eleventy doesn't crash. Instead, it remains in this weird limbo state where the watcher continues to run but builds cease.

The fix has two parts:

1. Awaiting the `setTimeout` so that `try/catch` works
2. Making sure that we catch non-fatal errors and mark the build as finished.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -4,6 +4,7 @@ const TemplateData = require("./TemplateData");
 const TemplateWriter = require("./TemplateWriter");
 const EleventyExtensionMap = require("./EleventyExtensionMap");
 const EleventyErrorHandler = require("./EleventyErrorHandler");
+const EleventyBaseError = require("./EleventyBaseError");
 const EleventyServe = require("./EleventyServe");
 const EleventyWatch = require("./EleventyWatch");
 const EleventyWatchTargets = require("./EleventyWatchTargets");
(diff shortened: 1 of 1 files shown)


zachleat:
Shipping with 1.0 thank you!

PR #1079, opened 2020-04-08
fix silent crashes in 11ty

### LINKED ISSUE #1081: 11ty crashes silently: doesn't output any files while not reporting any error.
**Describe the bug**

Some user configuration and developer mistakes, which cause the Template parts of the code to throw an error and make 11ty fail, are silently eaten and ignored. 

Not even having the DEBUG output as per https://www.11ty.dev/docs/debugging/ was any help: the failure goes unmentioned there as well. When you start to trace through the code you'll discover that there's some inexplicable *lack* of subsequent `debug(...)` output for several input files.

The result then is 11ty reporting success and you staring at a completely empty output directory and going *WTF*?

**To Reproduce**

See #1080 for a deliberately introduced failure (an exception which is rethrown) which showcases the issue: without #1079, 11ty will crash silently thanks to a Promise without `catch` way up the call chain. (See also the #1079 and #1080 patch diffs.)

**Expected behavior**

The exception (**any fatal failure** in that code tree anyway) should have been caught an reported, so that the user has something to read and wonder about while the output is not generated due to user and/or dev error.

**Environment:**

- Win 10 / node 12 
- Eleventy Version: both latest stable `npm i` version (0.10.0) as per today *and* also observed in the `master` branch commit which I forked the showcase and fix pullreqs off: #1079 and #1080 

**Additional context**

Ran into this while going through the code as I attempted to dig up what was going wrong with the local rig.

This kept me from discovering that at least one of my problems was some obscure stuff badly failing around permalinks as defined in a `_data/layouts/template.11ty.js` JavaScript layout template (following the spirit of https://www.11ty.dev/docs/languages/javascript/#permalink-function)

GerHobbelt:
fix silent crashes in 11ty: any failures with fatal consequences for the build process MUST be reported to the user for diagnosis.

Fixes #1081.

### DIFF
--- a/src/TemplateWriter.js
+++ b/src/TemplateWriter.js
@@ -153,10 +153,9 @@ class TemplateWriter {
     for (let path of paths) {
       if (this.extensionMap.hasEngine(path)) {
         promises.push(
-          this.templateMap.add(this._createTemplate(path)).then(() => {
-            debug(`${path} added to map.`);
-          })
+          await this.templateMap.add(this._createTemplate(path))
         );
+        debug(`${path} added to map.`);
       }
     }
 


MadeByMike:
If someone can tell me what causes the crash I'll write a test that confirms this. 

Also correct me if wrong but the best way to deal with this might be to put a catch on `promise.all(promises)` on line [63](https://github.com/11ty/eleventy/pull/1079/files#diff-904664a84856a301e63ddd05faa65cbbL163) we can then throw with the appropriate errors if there are multiple?

GerHobbelt:
See #1080: I did that one back then to showcase the crash.

zachleat:
Yeah we can’t await there as it’d bottleneck the loop.

zachleat:
Pushed something similar to this!

PR #1080, opened 2020-04-08
Showcase silent crash: this demos the type of crash fixed with pullreq #1079

GerHobbelt:
This crux is the rethrow (`throw ex;`) in there: such a deliberate exception (or any other failure in this code) should cause the app-level registered exception handler to kick in but no such luck.

The fix is #1079

# WARNING:

do NOT merge this pullreq. It's only there to showcase the issue.

### DIFF
--- a/src/TemplateLayout.js
+++ b/src/TemplateLayout.js
@@ -43,9 +43,15 @@ class TemplateLayout extends TemplateContent {
     }
 
     try {
-      throw new Error("arg")
-    } catch (ex) { console.error("Showcase silent crashes inside 11ty: this one will NOT b0rk the run!", ex); /* silently crashes... */}
-    
+      throw new Error("arg");
+    } catch (ex) {
+      console.error(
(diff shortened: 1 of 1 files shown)


Ryuno-Ki:
> do NOT merge this pullreq. It's only there to showcase the issue.

Hm, GitLab has the feature, that if you prepend a `WIP: ` to the PR title, it will be flagged as non-mergeable. Not sure, whether GitHub behaves the same way by now …

stevenpetryk:
@Ryuno-Ki you can open a ["draft PR"](https://github.blog/2019-02-14-introducing-draft-pull-requests/) that can't be merged unless the author marks it as "ready for review".

Ryuno-Ki:
Ah, nice one! Must have missed the news announcing this, so thanks for bringing it to my attention.

zachleat:
Confirmed this is fixed and errors appropriately 
image

PR #1083, opened 2020-04-09
Improve user feedback when 11ty reports 'Unsupported engine'

GerHobbelt:
List the supported engine extensions too as part of the error report, when you happen to feed 11ty an unsupported layout template file, e.g. `_data/layout/default.js` (note the lack of `.11ty` in that filename).

As also mentioned in #1081 this one had me thwarted for quite a while until I found out I had missed the **mandatory** `11ty` in `11ty.js` JavaScript Layout Template file naming -- it didn't help then that I hadn't found #1079 yet: once 11ty began to b0rk properly, this patch was the bit that would have sped me along the way a tad faster.


zachleat:
Thank you!

Issue #1125, opened 2020-04-27
Create a named instance of the browser-sync server


hirusi:
**Is your feature request related to a problem? Please describe.**
I have a starter-kit geared towards eleventy. I was using my own browserSync server to get full control over when to trigger a reload (for example, after my `css` or `img` tasks finish). I recently switched to the internal browserSync server for various reasons - chief among them getting ready in anticipation of incremental builds.

eleventy is internally using `browserSync` as well and [allows specifying options](https://www.11ty.dev/docs/config/#override-browsersync-server-options).
However, the instance created is not a named instance. It is therefore not possible for my tasks to look up if there is any active server and then hook into it.

**Describe the solution you'd like**
This should be a one line change here by passing a name string to the `create()` method.

https://github.com/11ty/eleventy/blob/master/src/EleventyServe.js#L119

This will then allow one to [get the instance of the server](https://www.browsersync.io/docs/api#api-get) and trigger page reloads when necessary.

I'll open a PR shortly.

**Describe alternatives you've considered**
I have looked through the [browserSync options](https://www.browsersync.io/docs/options) as suggested in the docs. It is not possible to configure a name of the server through options.

**Additional context**
I use `gulp` to create an asset pipeline and require the server to reload when my tasks finish. These tasks don't affect eleventy's built files.

Ryuno-Ki:
Do you have suggestions for the name?

hirusi:
> Do you have suggestions for the name?
The name can be anything, really - I've gone for `eleventyServer` in my PR.

> Are there limitations or considerations to take into account?
Not that I'm aware of, sorry.

hirusi:
This seems redundant as a replacement is being explored in #1305.

hirusi:
Sorry - reopened as it seems to be in preliminary stages. If and when a replacement is chosen, we can close this.

Either way, it's probably good to have linked to the issue so it is easily accessible in the context of the replacement.

zachleat:
Agree! This is a great addition—thank you! The named instance is `eleventy-server`

zachleat:
Fixed by #1126

PR #1126, opened 2020-04-27
Named broswer-sync instance

hirusi:
Solves #1125.


zachleat:
Merging this although I think I would prefer the name of the server to use a dash instead `eleventy-server` (sorry for this 😅).

Thank you!

PR #1161, opened 2020-05-09
Allow to configure an object tree for the permalink key

deiv:
This permits to put the permalink key outside of data root instead
to search the key only as a field at the root of the data object.

The use case, for me, is that I want to override the permalink value globally.

### DIFF
--- a/src/Plugins/Pagination.js
+++ b/src/Plugins/Pagination.js
@@ -182,7 +182,7 @@ class Pagination {
       let cloned = tmpl.clone();
 
       // TODO maybe also move this permalink additions up into the pagination class
-      if (pageNumber > 0 && !this.data[this.config.keys.permalink]) {
+      if (pageNumber > 0 && !lodashGet(this.data, this.config.keys.permalink)) {
         cloned.setExtraOutputSubdirectory(pageNumber);
       }
 
--- a/src/Template.js
+++ b/src/Template.js
@@ -2,6 +2,7 @@ const fs = require("fs-extra");
 const parsePath = require("parse-filepath");
 const normalize = require("normalize-path");
 const lodashIsObject = require("lodash/isObject");
+const lodashGet = require("lodash/get");
 const { DateTime } = require("luxon");
 
 const TemplateData = require("./TemplateData");
@@ -119,7 +120,7 @@ class Template extends TemplateContent {
       data = await this.getData();
     }
 
-    let permalink = data[this.config.keys.permalink];
+    let permalink = lodashGet(data, this.config.keys.permalink);
     if (permalink) {
       // render variables inside permalink front matter, bypass markdown
       let permalinkValue;


MadeByMike:
TBH, I don't understand what this is doing.

deiv:
@MadeByMike 

Well, my need for this, is that I want to change the default value of the permalink field globally, so instead of using the defaults for it. that generates the next paths:

> Input File -> _input/subdir/template.njk
> Output File -> _site/subdir/template/index.html

I want a "plain structure", for example:

> Input File ->_input/subdir/template.njk
> Output File -> _site/subdir/template.html


I see that it could be done in the template front matter data, for example:

```yaml
---
permalink: "this-is-a-new-path/subdirectory/testing/index.html"
---
```

But I don't want to add a permalink key in every template (or create a global layout to add the permalink key). I need to override the default value for the permalink globally.

---

My first aproach was to override it using a "global data file" [0], but i was unable to find a way to override it. The permalink field by default is at the root of the data structure, so I can't find a way to override something at the root.

For example this is the json data that a template receives:

```
{
  title: 'post1',
  description: 'post1 description',
  date: 2018-05-01T00:00:00.000Z,
  tags: [ 'another-tag' ],
  eleventyNavigation: {
      key: '2post-2-1',
      parent: '2post-2'
  },
  permalink: '{{ filePathStem }}.md',
  page: {
      date: 2018-05-01T00:00: 00.000Z,
      inputPath: './content/input/post/post1.md',
      fileSlug: 'post1',
      filePathStem: '/post/post1',
      url: '/.md'
    },
   collections: {}
}
```

If I add the next "_data/addedglobaldata.json";

```json
[
  "permalink":  "new permalink"
]
```
I get:

```
{
  title: 'post1',
  description: 'post1 description',
  date: 2018-05-01T00:00:00.000Z,
  tags: [ 'another-tag' ],
  eleventyNavigation: {
      key: '2post-2-1',
      parent: '2post-2'
  },
  permalink: '{{ filePathStem }}.md',
  page: {
      date: 2018-05-01T00:00: 00.000Z,
      inputPath: './content/input/post/post1.md',
      fileSlug: 'post1',
      filePathStem: '/post/post1',
      url: '/.md'
    },
   collections: {},
   addedglobaldata: {    <-- '_data/addedglobaldata.json' added here 
      permalink:   "new permalink"
   }
}
```

Maybe I miss something, but there's not an special file to allow override things at the template data json root level.


---

The next aproach was to add at the eleventy config file ".eleventy.js", an override for the permalink key. Example:

```javascript
    return {
        templateFormats: [
            "md"
        ],
        keys: {
            permalink: "customconfig.permalink" <-- override for the key to look for the permalink
        },
    };
};
```
So, if I add a global data file named "_data/customconfig.json" like this;

```json
[
  "permalink":  "new permalink"
]
```

I was thinking that the value for the permalink when the template is generated, is obtained from the new key "customconfig.permalink". The template will receive the next data:

```
{
  title: 'post1',
  description: 'post1 description',
  date: 2018-05-01T00:00:00.000Z,
  tags: [ 'another-tag' ],
  eleventyNavigation: {
      key: '2post-2-1',
      parent: '2post-2'
  },
  permalink: '{{ filePathStem }}.md',
  page: {
      date: 2018-05-01T00:00: 00.000Z,
      inputPath: './content/input/post/post1.md',
      fileSlug: 'post1',
      filePathStem: '/post/post1',
      url: '/.md'
    },
   collections: {},
   customconfig: {
      permalink:   "new permalink" <- when the tempalte is generated, it uses this key (customconfig.permalink) to get the permalink 
   }
}
```

But because the code that gets the permalink key value uses an array indexing over the data object:

`let permalink = data[this.config.keys.permalink];`

It looks for a field named "customconfig.permalink" at the root of the structure, because it not traverses the structure correctly.

The dot notation at "customconfig.permalink", doesn't gets interpreted correctly as I expected. I expect that the permalink value is extracted from the field "permalink" inside the "customconfig" object. Example:

Expected:

```
{
  title: 'post1',
  description: 'post1 description',
  date: 2018-05-01T00:00:00.000Z,
  tags: [ 'another-tag' ],
  eleventyNavigation: {
      key: '2post-2-1',
      parent: '2post-2'
  },
  permalink: '{{ filePathStem }}.md',
  page: {
      date: 2018-05-01T00:00: 00.000Z,
      inputPath: './content/input/post/post1.md',
      fileSlug: 'post1',
      filePathStem: '/post/post1',
      url: '/.md'
    },
   collections: {},
   customconfig: {
      permalink:   "new permalink" <- get the permalink from here !
   }
}
```

---

That's why this changes proposes to use lodash/get [1] to retrieve the value of the permalink. instead of using a simple array indexing. lodash/get will traverse the object path correctly.

---

[0] <https://www.11ty.dev/docs/data-global/>
[1] <https://lodash.com/docs/4.17.15#get>

MadeByMike:
Index files are a pretty good default. It means URLs work as expected on most static servers. 

To use your example of the output: `/subdir/template.html` - this means the following URL will work: `/subdir/template.html` and `/subdir/template` too, but maybe NOT: `/subdir/template/`. If you end up adding more pages in `/subdir/template/` e.g: `/subdir/template/index.html`, the URLs `/subdir/template` and `/subdir/template/` will serve different files. Which could be confusing. Even if you avoid index files all together this means generally URLs will be less intuitive. Of course this all depends on the server configuration.

Assuming you know all the above and want to do it anyway, if you override the permalink globally won't all your pages write to the same file location, overwriting each other?

The only way I can see to do this without conflict is to specify this for each page, which as you mentioned, the mechanism for this is the `permalink` value in front-matter.

What _might_ be generally more useful and not require config on each page is a function to rewrite paths. Kinda like transforms, but simply a function that takes the output path and data and allows you to return a different output path.  Something like:

```
const transformPath = ({inputPath, outputPath, data}) => {
  if (outputPath.endsWith("/index.html")) {
    const fileName = inputPath.substring(inputPath.lastIndexOf('/')+1).split(".")[0] 
    return outputPath.replace("index", fileName)
  }
  return outputPath;
}
```
(This code works for as a simple example case - but so many potential edge cases)

I could potentially see more uses for this than a static config based solution.

deiv:
> Index files are a pretty good default. It means URLs work as expected on most static servers.
> 
> To use your example of the output: `/subdir/template.html` - this means the following URL will work: `/subdir/template.html` and `/subdir/template` too, but maybe NOT: `/subdir/template/`. If you end up adding more pages in `/subdir/template/` e.g: `/subdir/template/index.html`, the URLs `/subdir/template` and `/subdir/template/` will serve different files. Which could be confusing. Even if you avoid index files all together this means generally URLs will be less intuitive. Of course this all depends on the server configuration.

The page strucuture that I want, in this case, is same that:

- index.html
- fruits.html
  + fruits/apple.html
  + fruits/orange.html
- cars.html
  + cars/ford.html
  + cars/fiat.html


> Assuming you know all the above and want to do it anyway, if you override the permalink globally won't all your pages write to the same file location, overwriting each other?

Not, if you use javascript data file, instead of yaml. For the same structure that describe prior, I added the next (the output page strucuture, render the same as the input):

input/_data/customconfig.js:
```javascript
module.exports = {
  "permalink": "{{ page.filePathStem }}.html"
}
```

This way, the global permalink, is interpreted when each page is about to be rendered (the latters), instead of when the customconfig,js file is parsed. So no overwriting occurs.


> The only way I can see to do this without conflict is to specify this for each page, which as you mentioned, the mechanism for this is the `permalink` value in front-matter.
> 
> What _might_ be generally more useful and not require config on each page is a function to rewrite paths. Kinda like transforms, but simply a function that takes the output path and data and allows you to return a different output path. Something like:
> 
> ```
> const transformPath = ({inputPath, outputPath, data}) => {
>   if (outputPath.endsWith("/index.html")) {
>     const fileName = inputPath.substring(inputPath.lastIndexOf('/')+1).split(".")[0] 
>     return outputPath.replace("index", fileName)
>   }
>   return outputPath;
> }
> ```
> 
> (This code works as a simple example case - but so many potential edge cases)
> 
> I could potentially see more uses for this than a static config based solution.

Well, this is something similar to adding the permalink in a global layout. It adds another step to calculate the permalink, and another "file/piece of code", to simply allow to change the permalink globally.

Allowing to override the permalink directly, we don't add any extra processing step to the final permalink value. This is one of the goals of this patch (secondary). And to reduce the configuration/code needed for the permalink overriding.

deiv:
By the way:

This "maybe" could be done with the new feature you added (Add globalData via config - https://github.com/11ty/eleventy/pull/1280), instead of renaming the permalink key (need to take a look).

Now, with this new `addGlobalData` method, is not needed to make the workaround I describe before, to allow to change/adding global data.

But it could be good too, to use lodashGet to get the permalink key, if the code allows to override it.

MadeByMike:
addGlobal data shouldn't be able to do anything you can't do with a data file.

Can I ask, why are you keen on the particular page structure you mentioned?:

```
index.html
fruits.html
  fruits/apple.html
  fruits/orange.html
cars.html
  cars/ford.html
  cars/fiat.html
```

You get practically the same results for users either way.

deiv:
The generated pages will be served by a rest service. Im using eleventy as a layer (the big one), in a knowledge content manager (note taking) app. So i need a page structure that reemsembles a tree.

deiv:
Hi !

After this good conversation. Do you have any intention on merge it ?

zachleat:
I didn’t have a read of all the comments here but I do like this idea. However, I also think having this inconsistency with the other configurable keys will be confusing. See https://github.com/11ty/eleventy/blob/1d7a96737912ff93f7d149a648419b84c9012a7e/src/defaultConfig.js#L42

deiv:
> I didn’t have a read of all the comments here but I do like this idea. However, I also think having this inconsistency with the other configurable keys will be confusing. See
> 
>https://github.com/11ty/eleventy/blob/1d7a96737912ff93f7d149a648419b84c9012a7e/src/defaultConfig.js#L42

Uhm...

 so you mean that al the configurable keys should allow to configure them outside the root object? Same that permalink in this patch?

zachleat:
Going through some old PRs here and my initial reaction on re-reading is that the original use case is supported fine with global data files—use a `_data/permalink.js` file to do this? Also I believe `eleventyConfig.addGlobalData` would work too?

Although I’m guessing there will be edge cases (bugs) if your project uses a bunch of different template syntaxes and the template syntax is inconsistent with what you set in those global data files.

I think you’ll have better luck with Eleventy Computed Data. You can set `permalink` in there just fine. https://www.11ty.dev/docs/data-computed/

I’m going to close this for now—sorry. Please let me know if Computed Data doesn’t solve this use case.

PR #1196, opened 2020-05-19
chore: update minimum Node.js LTS version to v10 rather than the EOL v8

bnb:
Currently, the minimum supported version is Node.js v8. Node.js v8 went EOL on December 31, 2019.

This PR updates both `engines` to check for at least Node.js 10 and the CI to only test on 10 and 12 (14 should be added at some point, since it'll be going LTS in ~October).

It seems that the docs may need to be updated, too.

I searched open PRs and couldn't find anything. Feel free to close this if it's unwelcome 🙇🏻‍♂️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
-  - 8
   - 10
   - 12
+  - 14
 before_script:
   - npm install
 script: npm run test
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "main": "src/Eleventy.js",
   "license": "MIT",
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
   },
   "funding": {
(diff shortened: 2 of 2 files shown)


MadeByMike:
"and the CI to only test on 10 and 12 (14 should be added at some point, since it'll be going LTS in ~October)."

@bnb - This looks great but I can only see change to `engines` were there meant to be CI changes in this PR?

Also just curious, does updating engines stop Eleventy from working on older Node environments? We should possibly still updated it since 8 is EOL but if it's going to break some builds I think we'd need to be clear about it and maybe offer a workaround for them?

bnb:
Fixed 👍

zachleat:
Yeah we’re going to retire support for Node 8 in Eleventy 1.0 so it shouldn’t break any existing builds.

zachleat:
Thanks for this!

PR #1209, opened 2020-05-24
Pagination: pass metadata to before callback

vseventer:
This PR adds a second argument for passing page data in the paginations' `before` callback.

This is useful (especially when leveraging the Data Cascade, where a piece of metadata might be elsewhere in the cascade) when you want to process (filter or modify) based on a value in the frontmatter:
```js
---js
{
  keyword: "item3",
  pagination: {
    data: "items",
    size: 1,
    before: function(data, metadata) {
      return data.filter(el => el === metadata.keyword);
    }
  },
  items: ["item1", "item2", "item3", "item4", "item5", "item6"]
}
---
{{ pagination.items[0] }} <- item3

```

### DIFF
--- a/src/Plugins/Pagination.js
+++ b/src/Plugins/Pagination.js
@@ -120,7 +120,7 @@ class Pagination {
       typeof this.data.pagination.before === "function"
     ) {
       // we don’t need to make a copy of this because we already .filter() above
-      result = this.data.pagination.before(result);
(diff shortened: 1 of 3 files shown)


zachleat:
Some commentary at #1522

zachleat:
Shipping with 1.0.1, earliest canary is `1.0.1-canary.4`

PR #1276, opened 2020-06-21
Add support for relative protocols in url filter

### LINKED ISSUE #1247: url filter should detect absolute urls without the protocol
I'm using the `url` filter for images in this way:

```njk
<img src="{{ image.src | url }}">
```

But sometimes, the images are external urls. To make them work in http and https contexts, I remove the protocol, so the urls are like this: `//placehold.it/300x100`. Seems like the filter does not detect this and generate the invalid url `/placehold.it/300x100`.

I can see that the filter can detect http and https protocols (https://github.com/11ty/eleventy/blob/master/src/Filters/Url.js#L8) so I propose adding another condition: `url.indexOf("//") === 0`.

oscarotero:
This PR fixes #1247

Currently, the url filter normalize values starting with `//` as if they where paths, so for example a url to `//example.com` is changed to `/example.com` (or `pathPrefix/example.com` if there's a path prefix configured).

But technically, `//example.com` is a perfectly valid url that should be resolved to `http://example.com` or `https://example.com`, depending of the protocol of the current document location.

This PR fixes this problem. Note that this is a breaking change.


MadeByMike:
This looks good to me. Although it's also potentially a breaking change for some people. I think we'd need to document and communicate this change thoroughly. 

@zachleat this is slightly off topic but it would be super useful if the website documentation was in the same repo as eleventy (possibility even a monorepo (...I know, I know, but you can do it with a light touch, it's not all bad!)) 

I also think using changeset could be a good addition to managing things: https://github.com/atlassian/changesets.

Let me know if you'd be open to PRs for either of these.

src/Filters/Url.js
```diff
@@ -1,14 +1,15 @@
...
+module.exports = function (url, pathPrefix) {
   // work with undefined
   url = url || "";
 
   if (
     validUrl.isUri(url) ||
     url.indexOf("http://") === 0 ||
-    url.indexOf("https://") === 0
+    url.indexOf("https://") === 0 ||
+    (url.indexOf("//") === 0 && url !== "//")
```
dgrammatiko, line 12: This could also be written as `(url.startsWith("//") && url !== "//")` 
Actually all of them in the conditional can be replaced with `startsWith()` as there is no way the url to contain the protocol in any other position except the... start of the string

nhoizey:
I think it should be mentioned that protocol relative URLs can be harmful, and are now considered an anti-pattern:
https://www.paulirish.com/2010/the-protocol-relative-url/

zachleat:
This will ship with 1.0

PR #1306, opened 2020-07-07
Add configuration to support serve middlewares.

calavera:
This allows people to configure server middlewares for serving different content besides static assets. The middlewares are expected to be compatible with http-proxy-middleware, that's fully supported by browserSync.

@zachleat I'm not super familiar with the codebase, and some logic might be better handled in a different way, so please feel free to make suggestions.

I can add docs with examples on how to use this, but I was not sure what the best place was.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/src/EleventyServe.js
+++ b/src/EleventyServe.js
@@ -36,7 +36,7 @@ class EleventyServe {
     return TemplatePath.join(this.getRedirectDir(dirName), "index.html");
   }
 
-  getOptions(port) {
+  async getOptions(port) {
     let pathPrefix = this.getPathPrefix();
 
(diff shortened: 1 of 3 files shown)


calavera:
Now that I look at the changes, I'm starting to think that it could be more simple to implement if the middlewares are just the expected callbacks to add to the server, instead of calling a callback to configure them that returns the middleware callback.

calavera:
And I just realized that I can pretty much do the same thing I want without changes in the framework, with `setBrowserSyncConfig`, so I'm going to close this.

PR #1317, opened 2020-07-16
Add changed file path in beforeWatch event args

dafiulh:
So, this will allows people to know which files changed when the `beforeWatch` event occurred, for example:
```js
eleventyConfig.on('beforeWatch', (path) => {
    if (path.endsWith(".html")) {
        // do something
    }
 });
```

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -452,12 +452,12 @@ Arguments:
    * @private
    * @method
    */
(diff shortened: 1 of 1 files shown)


zachleat:
Yes absolutely this is coming.

zachleat:
Oops sorry this was a PR (yes dev work for this is in)

PR #1320, opened 2020-07-19
Pass the `dir` config options to plugins

MadeByMike:
I've found numerous cases where knowing the input and output dir is required for plugins.

Because of the execution of merging options it was not as simple as first imagined. I passed the responsibility for running plugins to the TemplateConfig class. This felt similar to a lot of other config options which are set and run later.

I'm looking for feedback on whether the raw config values should be passed as I've done here... or if we should resolve these further first? Also... should we pass anything else from TemplateConfig to plugins?


zachleat:
Thank you! Only a year later 👀

PR #1403, opened 2020-09-08
Add config option to setEleventyIgnoreFile

scottishstoater:
Eleventy Config option to define what `.eleventyignore` file users would like to use. The reference to the `.eleventyignore` file is currently hardcoded, adding this options makes it slightly more flexible while maintaining the default with zero-config.  

This will allow people to define what files Eleventy ignores based on environments it's running in. [See initial Twitter discussion here for a bit of background](https://twitter.com/scottishstoater/status/1302914523788173312).

## Example use case

A project has a folder with 10,000 markdown articles that all share a single template. The developer would like to speed up development by defining a development set of 10-20 markdown articles that represents the larger folder on a smaller scale. 

## Problem 

At present there is no way to tell Eleventy to ignore files based on the environment it's running. 

## Solution 

Allow developers to define a custom ignore file within the EleventyConfig.

## Example Solution

_.dev.eleventyignore_
```
src/_articles/
```

_.eleventyignore_
```
src/_example_articles/
```

_.eleventy.js_
```js
const isProduction = process.env.ELEVENTY_ENV === 'production';

    ...

    if(!isProduction) {
        eleventyConfig.setEleventyIgnoreFile(".dev.eleventyignore");
    }

    ...
};
```

### DIFF
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -271,11 +271,11 @@ class EleventyFiles {
 
     if (this.config.eleventyignoreOverride !== false) {
       let eleventyIgnores = [
-        TemplatePath.join(rootDirectory, ".eleventyignore")
+        TemplatePath.join(rootDirectory, this.config.eleventyIgnoreFile)
       ];
       if (rootDirectory !== absoluteInputDir) {
         eleventyIgnores.push(
-          TemplatePath.join(this.inputDir, ".eleventyignore")
+          TemplatePath.join(this.inputDir, this.config.eleventyIgnoreFile)
         );
       }
 
--- a/src/UserConfig.js
+++ b/src/UserConfig.js
@@ -384,6 +384,10 @@ class UserConfig {
     this.useGitIgnore = !!enabled;
   }
 
+  setEleventyIgnoreFile(file) {
+    this.eleventyIgnoreFile = file;
+  }
+
   addShortcode(name, callback) {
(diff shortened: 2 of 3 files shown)


scottishstoater:
ahh, just seeing the checks failed. I'll look into that.

binyamin:
I would rather forget any custom ignore files, and have the eleventyConfig accept an array of globs to ignore. You could dynamically set the config option based on the environment, without adding extra files.

edwardhorsford:
Personally, I'd love to be able to set ignore files dynamically. Until 'drafts' as a feature are supported, I'm left with hardcoding files to ignore in my ignore files, and commenting them out when I want to see them.

If it could be set in my `eleventyConfig` I could dynamically set which files to ignore.

scottishstoater:
@binyamin I totally agree with you. Dynamically creating the ignore glob array within the eleventyConfig rather than an additional hardcoded file sounds like a much more flexible solution. 

I actually stumbled across a config option within the [EleventyFiles](https://github.com/11ty/eleventy/blob/master/src/EleventyFiles.js#L272) class called `eleventyignoreOverride` which looks like it takes a glob string. It's only referenced twice within the source code and again in the tests. 

I can't get my head around where it's being set and I'm not entirely sure what it does but it sounds like it could be lurking in the same area as we are exploring. Maybe @zachleat could shed some light on it. 

🙂

Issue #631, opened 2019-07-26
Show warning when data files override each other


zachleat:
e.g. `mydatafile.json`, `mydatafile.11tydata.js`.

Order is documented here: https://www.11ty.io/docs/data-template-dir/

via https://github.com/11ty/eleventy/pull/624#issuecomment-515587598

zachleat:
> This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

> View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

zachleat:
Shipping with 2.0.0-canary.19.

PR #1462, opened 2020-10-20
Warning when data files override each other

Obayanju:
Implements #631 

Shows warning when template, directory, and parent directory data files override.
I wasn't sure if we need to add tests for warnings?


zachleat:
Looks great, thank you! This will ship with 2.0.0-canary.19.

PR #1475, opened 2020-10-26
Add Node 15 to CI

valtlai:
Node 15 is now released and [officially supported](https://nodejs.org/en/about/releases/) Node version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        node: ["10", "12", "14"]
+        node: ["10", "12", "14", "15"]
     name: Node.js ${{ matrix.node }} on ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


zachleat:
I’ve been kind of avoiding odd numbered nodes for reasons outlined in this article: https://tamalweb.com/which-nodejs-version

Should I reconsider this? It seems to have worked out okay so far

valtlai:
I’ve used the odd-numbered Node versions in my personal website without problems. I’ve considered the LTS versions being made for enterprise use and the *current* version being suitable for most of the users – or hobbyist, at least. I know the non-LTS versions are generally considered somewhat experimental or beta-like.

However I think it would be good to ensure Eleventy is working also on the odd-numbered Node versions. On the other hand, I understand it may not be worth of it to support a Node version that’ll become deprecated after six months.

zachleat:
Right on—I’m okay to merge this and reserve the right to change my mind again later 😅

zachleat:
Thanks!

valtlai:
> Right on—I’m okay to merge this and reserve the right to change my mind again later 😅

That’s very okay. 😁 Thank you for merging!

Issue #1510, opened 2020-11-12
Include inputPath in environment options passed to markdown


GeorgeNance:
**Is your feature request related to a problem? Please describe.**
Yes, It would be nice to have context to which file is being rendered when I am writing custom plugins for my markdown. 

I am trying to process images that are in the same folder as a markdown file.

**Describe the solution you'd like**

Simply passing the inputPath to `mdlib.render` in Markdown.js

**Describe alternatives you've considered**
There is no easy alternate at this point. 

**Additional context**
N/A

zachleat:
Hmm—does `page` solve this? `page.inputPath`
https://www.11ty.dev/docs/data-eleventy-supplied/#page-variable-contents

GeorgeNance:
I don’t believe we are able to access page variables from plugins that are called when rendering . I may be wrong .

zachleat:
> This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

> View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

manunamz:
For others who also want to access `inputPath` in a markdown-it plugin:

This is already possible, if a bit loopy. See details below.

- 11ty stuffs its `data` into the markdown-it library's render function [here](https://github.com/11ty/eleventy/blob/master/src/Engines/Markdown.js#L85).
- Markdown-it's `render` function receives it [here](https://github.com/markdown-it/markdown-it/blob/df4607f1d4d4be7fdc32e71c04109aea8cc373fa/lib/index.js#L541), with [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cd29f1c657713344b002f8a87de631cb3e60989a/types/markdown-it/lib/index.d.ts#L298) type declaration.
- You may then access `env.page.inputPath` (among other 11ty supplied data) inside of your render rules, which might look something like:

```js
function customPlugin (md, opts) {
  md.renderer.rules.custom_render_rule = function (tokens, index, opts, env) {
    // do some things...
    return `now you can access inputPath! ${env.page.inputPath}`
  }
}
```

### However...

[...]

PR #1511, opened 2020-11-12
Added inputPath as an environment option for markdown engine

GeorgeNance:
Solves #1510

Adding the inputPath to the markdown renderer so markdown it can be referenced by plugins.

### DIFF
--- a/src/Engines/Markdown.js
+++ b/src/Engines/Markdown.js
@@ -71,7 +71,7 @@ class Markdown extends TemplateEngine {
         return async function (data) {
           let fn = await fnReady;
           let preTemplateEngineRender = await fn(data);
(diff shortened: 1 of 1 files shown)


GeorgeNance:
If this is not a valid approach, maybe I can suggest adding a global for the current file being rendered ?

zachleat:
The change seems to break two tests:
image

zachleat:
> This is an automated message to let you know that this issue requires additional input from the original poster and for the health of the repository issue tracker the issue will be closed. This is to help alleviate issues hanging open waiting for a response from the original poster.

> Do not let the issue’s closing deter you. After you’ve supplied the required information, the issue will be reopened. Thanks!

PR #1537, opened 2020-12-03
Make cachePartialFiles and getPartials async

NotWoods:
The only remaining sync call to `getPartials` was in the constructor of `Handlebars`. Moving the partial registration to `compile` lets us make the other functions async!

`registerPartial` can handle multiple partials, so the for loop has been removed.

I added a few JSDoc comments to help me follow the code. I can remove this if desired.


src/Engines/TemplateEngine.js
```diff
@@ -81,22 +85,22 @@ class TemplateEngine {
...
         );
-      });
+      }));
       bench.after();
     }
 
     partialFiles = TemplatePath.addLeadingDotSlashArray(partialFiles);
 
-    for (let j = 0, k = partialFiles.length; j < k; j++) {
+    await Promise.all(partialFiles.map(async (partialFile) => {
```
zachleat, line 101: ah, hmm—I’m curious about this `map` usage. You’re Promise.all-ing the return there but the map doesn’t seem to return a promise? Is there some additional efficiency we can get here?

src/Engines/TemplateEngine.js
```diff
@@ -151,6 +158,15 @@ class TemplateEngine {
   get defaultTemplateFileExtension() {
     return "html";
   }
+
+  /**
+   * Check whether the dairy product is solid at room temperature.
```
zachleat, line 163: 😇

zachleat (commented):
Howdy! This looks great but I did just have one small question

src/Engines/TemplateEngine.js
Ryuno-Ki: By using an async arrow function, the return value will be a Promise.

pdehaan: Could probably use `.forEach()` here since I don't see anything that is _returning_ a value within `.map()`. Same w/ L88 above.

Ryuno-Ki: It's returning `Promise.resolve(undefined)`.
You're _not consuming_ the return value of `Promise.all`.
I'd argue, that `Promise.all` requires an `Array<Promise<any>>` to work on unless you want to experience undefined behaviour.

NotWoods: As Ryuno-Ki said, the `.map` call is needed because async functions return a promise, which is awaited by `Promise.all`. `forEach` would cause the function to proceed before the promises are done.

I could rewrite this using `.then` chains instead of async functions so there is a clear return value if preferred.

Ryuno-Ki:
Heads up, a merge conflict arose.

zachleat:
it’s happeningggggggg…

v2.0.0-canary.19

PR #1556, opened 2020-12-16
Adds ondemand support for serve mode

samthor:
This adds a `--ondemand` flag that, when combined with `--watch`, only builds pages which are requested. (This is basically the idea behind [11ty-fast-dev](https://github.com/samthor/11ty-fast-dev)).

In our testing, this provides ~1.5x speedup on test Eleventy sites during development. If you have costly transforms or filters, it speeds it up even more (this is the case for developer.chrome.com, where we unfortunately do HTML parsing and rewriting as part of this step).

If running your templates (and their filters) or transforms has global side-effects and you reply on that during development, then this probably isn't a good idea.

Cons:
- I'm not entirely sure how well this interacts with incremental build
- You might want this behavior _without_ BrowserSync (in fact, developer.chrome.com probably does, but this feature just makes sense behind `--watch`).

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -79,8 +79,14 @@ class Eleventy {
      */
     this.formatsOverride = null;
 
+    /** @member {Boolean} */
+    this.isOnDemand = false;
+
+    /** @member {Object<string, Function>} */
+    this._onDemandBuild = {};
+
     /** @member {Object} - tbd. */
-    this.eleventyServe = new EleventyServe();
+    this.eleventyServe = new EleventyServe(this._requestBuild.bind(this));
 
     /** @member {String} - Holds the path to the input directory. */
     this.rawInput = input;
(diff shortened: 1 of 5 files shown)


src/Eleventy.js
```diff
@@ -759,9 +808,14 @@ Arguments:
     await this.config.events.emit("beforeBuild");
 
     try {
-      let promise = this.writer.write();
+      if (this.isOnDemand) {
+        this._onDemandBuild = await this.writer.prepareOnDemand();
+        debug(`Generated ${Object.keys(this._onDemandBuild).length} on-demand candidates…`);
+      } else {
+        let promise = this.writer.write();
+        ret = await promise;
```
dgrammatiko, line 816:
```suggestion
        ret = await  this.writer.write();
```

Ryuno-Ki:
Any updates?

samthor:
I'm waiting for input from @zachleat.

I _suspect_ his work on [#1629](https://github.com/11ty/eleventy/pull/1629/files) might make this sort of approach easier, but I'm not sure yet.

PR #1560, opened 2020-12-22
Add Prettier configuration file and npm "format" script

mariusschulz:
This PR adds a _.prettierrc.json_ file and formats all JS files in the _src_ folder with Prettier.

I noticed that the _prettier_ package is installed, but that there's no Prettier configuration file. With the _.prettierrc.json_ file in place and the _prettier_ package installed, editors should now use consistent formatting. This should reduce some noise created by formatting-only changes in PRs.

Note that I'm not attached to any specific Prettier configuration. I tried to pick the one that most closely resembles the state of the codebase today. If anyone has strong opinions on certain options (e.g. `arrowParens` or `trailingComma`), please let me know, happy to update!


zachleat:
Oh interesting. Prettier is applied using `lint-staged` I believe as a pre-push hook? I do like the idea that editors would pick it up though.

Can you revert the prettier formats to make it easier to merge? We can re-apply after merging

zachleat:
https://github.com/11ty/eleventy/blob/39afa802ac7495102be038e84cb648cc1aa4a22e/package.json#L70-L73

mariusschulz:
Sure, that sounds good to me. I updated this PR to only include the _.prettierrc_ file as well as the `format` npm script. I'll submit another PR to format the entire current codebase.

mariusschulz:
I submitted another PR to format all JS files in `src/` with Prettier: #1661

mariusschulz:
@zachleat Are we good to merge this? :)

.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
```
matijs, line 1: How about renaming this file to `.prettierrc.json` to get the correct syntax highlighting for free?

zachleat:
Mergin’, thanks!

PR #1661, opened 2021-02-27
Format JS files in src/ with Prettier

mariusschulz:
Format all JS files in `src/` with Prettier.

See #1560 for the discussion.


zachleat:
I re-applied these formats manually—thank you!

PR #1662, opened 2021-02-27
Implement proof-of-concept for @

mikestreety:
A very rough concept for implementing an `@` like prefix to paths, to make them relative to the `.eleventyconfig.js` file instead of the input.

Helps for when the layouts, includes and data folders are not children of the input dir

e.g.

```
module.exports = function (config) {
	return {
		dir: {
			input: '@/app/content',
			data: '@/app/data',
			includes: '@/app/includes',
			layouts: '@/app/layouts',
		}
	};
};
```

More details can be found in [this repo](https://github.com/mikestreety/11ty-quandary).

Would help with instances [like this, too](https://gitlab.com/streety-sites/behind-the-source/-/blob/master/.eleventy.js)

### DIFF
--- a/src/TemplateConfig.js
+++ b/src/TemplateConfig.js
@@ -239,6 +239,9 @@ class TemplateConfig {
     // Unique
     merged.templateFormats = lodashUniq(merged.templateFormats);
 
+	// Append config path location
+    merged.configPath = path;
+
     debug("Current configuration: %o", merged);
 
     return merged;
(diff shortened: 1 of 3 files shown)


zachleat:
(keeping the context https://twitter.com/mikestreety/status/1365757241152987136)

zachleat:
I think the other interesting point to keep in mind is that it would be nice if `layout:` declarations in the data cascade could use this as well!

via https://toolsforthought.rocks/@BentleyDavis/109526129946337560

PR #1691, opened 2021-03-14
TemplateData: allow non-object data (e.g. strings)

milahu:


zachleat:
A few things here, this will work fine for Global Data files but will _not_ work for template/directory data files. We don’t want to return literals in custom data extensions because that would blow away the entire data cascade for that template. I’m guessing we have some room for improvement around error messaging though! That said, I do consider this is an improvement! I’ll merge and work on the warnings/errors for template/directory data files.

zachleat:
Shipping with 2.0.0-canary.10

Issue #699, opened 2019-09-18
Let me transform pagination data using JS before it is paginated


zachleat:
A callback function to let me operate on the pagination data prior to pagination

zachleat:
This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

zachleat:
This feature unfortunately overlaps with https://www.11ty.io/docs/pagination/#how-to-reverse-a-data-set-prior-to-pagination and docs will need to reflect that

zachleat:
- [ ] make sure this works with `alias`.

zachleat:
Docs have been added!

PR #1697, opened 2021-03-17
Pagination: bind `this` to `before` callback

Jaza:
Provides access to all page data, via `this` in the `before` callback function for pagination. Useful, for example, for filtering pagination items based on the permalink of the current page.


Jaza:
Related to #699

Jaza:
I think #1711 is a better solution

PR #1698, opened 2021-03-18
Pagination: add pageOnEmptyData option

Jaza:
Currently, if the data to paginate is empty, then eleventy simple generates zero pages. But you might want to always generate at least one page, and if there's no data, then perhaps show a message such as "no entries found", where the paged data would normally appear. This adds a new option `pageOnEmptyData` to pagination, to make such behavior possible.


zachleat:
I like this! I wonder if we can brainstorm a better name?

`pageOnEmptyData`? `emptyDataGeneratePage`? `emptyDataPage`?

Jaza:
@zachleat sure, `pageOnEmptyData` works for me, I struggled to come up with `pageIfNoData` and wasn't completely happy with it anyway.

zachleat:
Merged this one (and #1698!) Went with `generatePageOnEmptyData`. Shipping with 2.0.0-canary.10

bsastregx:
@zachleat I love eleventy. working on my first project I came to realize that if no posts exists on a collection where pagination is present, no page will be created, which is understandable. However, I might have a link on my global menu that points to the first page of the collection. Now that link takes the user to an inexistent page `Cannot GET 'pageName' `, and me (the developer) didn't have any message/warning on the terminal. Don't you think that behavior is dangerous? 

Because of this I created an issue: https://github.com/11ty/eleventy/issues/2928

@pdehaan suggested me to use `generatePageOnEmptyData`. I liked that solution. However, for any eleventy developer that does not know about `generatePageOnEmptyData` this behavior will be a surprise.

Now that I know about `generatePageOnEmptyData` my suggestion is the following:
If there is pagination && the collection is empty && generatePageOnEmptyData is not present, warn the user on the terminal. This could avoid potential undesired urls with no html.

bsastregx:
Or maybe `generatePageOnEmptyData` should be true by default. To be false, developer should add `generatePageOnEmptyData: false` to front matter.

PR #1699, opened 2021-03-18
feat: index datafiles with constant basename

milahu:
support global / constant basename like `index` for [Directory Data File](https://www.11ty.dev/docs/data-template-dir/)
so when renaming directories, we dont have to rename the directory data files too

`.eleventy.js`
```js
  return {
    jsDataFileBase: 'index',
  };
};
```

`some/path/index.11tydata.json`
```json
{
  "tags": "someTag"
} 
```

`some/path/someFile.md`
```md
# hello
```

now `some/path/someFile.md` (and other files in `some/path/`) is tagged with `someTag`

todo: add docs and tests

### DIFF
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -485,7 +485,7 @@ class TemplateData {
     // data suffix
     paths.push(base + dataSuffix + ".js");
     paths.push(base + dataSuffix + ".cjs");
-    paths.push(base + dataSuffix + ".json");
+    paths.push(base + dataSuffix + ".json"); // default: .11tydata.json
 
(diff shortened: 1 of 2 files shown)


zachleat:
Gonna merge this one, thank you! I will probably tinker with the configuration API to enable this before shipping it but look for it in 2.0.0-canary.19!

zachleat:
Now known as `dataFileDirBaseNameOverride` and setting this option will override the default behavior of looking for  `directory/directory.*` in the data cascade and use only `directory/index.*` instead.

Issue #461, opened 2019-03-22
Core pagination support for double layered pagination


zachleat:
Related issue #332, something with a little bit more DX might be warranted since this has come up a few times now.

zachleat:
This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

distinctperspective:
I converted too. @zachleat  I can help you with that.

jeromecoupe:
My 2 cents on this double layered pagination things

jeromecoupe:
Yup, using the same king of approach, albeit using custom taxonomies instead of tags:

https://github.com/jeromecoupe/11ty-paginate-taxonomies

MWDelaney:
Expanding the above into a working example...

After much heartache, and many compromises, I have paginated dynamic tag pages for blogging working.

I made a pull request against ZeroPoint demonstrating it:
https://github.com/MWDelaney/ZeroPoint/pull/31

I'm sure this could be simplified, and pagination more directly modified, but it works!

dave-kennedy:
I'm confused by the status of this issue. It's pinned to the top of the issues list, so I would expect a canonical answer here. @zachleat can you please clarify? Is the recommended solution [this one](https://github.com/11ty/eleventy/issues/332#issuecomment-445236776)? If so, it might make more sense to pin that issue instead.

Issue #332, opened 2018-12-06
Double-layered pagination?


zellwk:
I'm creating a blog. Each post are tagged. I've created each tag with the [zero maintenance tag pages](https://www.11ty.io/docs/quicktips/tag-pages/) tip. 

But each tag can potentially contain lots of posts. Can Eleventy create paginated tag pages? 

EG: 

1. /tags/tag1/
2. /tags/tag1/page2
3. /tags/tag1/page3
4. /tags/tag2/
5. /tags/tag2/page2
6. /tags/tag3/

And so on.

jevets:
I don't think there's a straightforward way. There are a couple issues about this but I don't think a solution was found, without some custom modules/preprocessing.
https://github.com/11ty/eleventy/issues/294
https://github.com/11ty/eleventy/issues/308


`_includes` files can't use pagination in their own frontmatter.


**But there's this idea from the docs** (I haven't tried this approach for anything):
https://www.11ty.io/docs/permalinks/#ignore-the-output-directory

Basically, let Eleventy compile a file and output it to the `_includes` directory. 
Then your other file will `include` the compiled file.

If you give this a go and if it works, please let us know how it goes.

zachleat:
This is a fascinating idea—I’ve been mulling it over since you posted it.

I think you can do this, but you would have to flatten your custom collection to a single layer to do it.

Use custom collections: https://www.11ty.io/docs/collections/#advanced%3A-custom-filtering-and-sorting.

Here’s how it would work:

```
// note that this uses the lodash.chunk method, so you’ll have to require that
eleventyConfig.addCollection("doublePagination", function(collection) {
		// Get unique list of tags
		let tagSet = new Set();
		collection.getAllSorted().map(function(item) {
			if( "tags" in item.data ) {
				let tags = item.data.tags;

				// optionally filter things out before you iterate over?
				for (let tag of tags) {
					tagSet.add(tag);
				}

			}
		});

		// Get each item that matches the tag
		let paginationSize = 3;
		let tagMap = [];
		let tagArray = [...tagSet];
		for( let tagName of tagArray) {
			let tagItems = collection.getFilteredByTag(tagName);
			let pagedItems = lodashChunk(tagItems, paginationSize);
			// console.log( tagName, tagItems.length, pagedItems.length );
			for( let pageNumber = 0, max = pagedItems.length; pageNumber < max; pageNumber++) {
				tagMap.push({
					tagName: tagName,
					pageNumber: pageNumber,
[...]

PR #1711, opened 2021-03-26
Pagination: let data key be a function

Jaza:
So that pagination data key can be determined dynamically,
based on a callback function that gets passed all template
data. So you can do something like:

    ---js
    {
      pagination: {
        data: (data) => {
          return "collections." + data.foo;
        },
    ...


Jaza:
I just stumbled across https://www.webstoemp.com/blog/basic-custom-taxonomies-with-eleventy/ , which led me to #332 , which I see has a multitude of other issues linked to it.

In case I didn't make it clear before (I don't think I did): this PR solves double pagination! (My other recently-opened PR #1697 also solves it, but I think this one is more elegant). That's exactly why I made this change - to make pagination work on tag pages.

@zachleat your code in https://github.com/11ty/eleventy/issues/332#issuecomment-445236776 is ok as an interim solution, but I think "double pagination" should be made possible in Eleventy, without having to copy-paste so much code (and you're kind of duplicates the chunking that the pagination plugin does internally).

So, this is potentially all we need, to fulfil the enhancement item #461 , which I see has quite a few votes, and is currently pinned.

Please consider getting this in!

Jaza:
@zachleat any chance you could take a look at this? Has been more than 3 months. There has been no other progress in resolving the double pagination problem, as far as I'm aware.

Jaza:
@zachleat please? This PR has been up for a whole year now. I've been running this change in production for that whole time too. Would you mind reviewing it?

Jaza:
Rebased against latest master and resolved conflicts

PR #1733, opened 2021-04-10
Add keyword arguments for Liquid shortcodes.

### LINKED ISSUE #1263: Support keyword arguments for Liquid shortcodes.

dawaltconley:
Fixes #1263. This removes `moo` and switches the argument lexer/parser for Liquid shortcodes to [liquid-args](https://www.npmjs.com/package/liquid-args), to support keyword arguments with output mirroring Nunjucks kwargs.

This includes some breaking changes in how Liquid shortcodes handle the whitespace in between arguments. With `liquid-args`, whitespace can be any number of space, tab, newline, or carriage return characters (`[ \t\n\r]`), and will allow a comma at the beginning of the whitespace. That means that the following tags will parse correctly:
```
{% shortcode arg1, arg2, arg3 %}
{% shortcode arg1 arg2 arg3 %}
{% shortcode arg1,arg2,arg3 %}
{% shortcode arg1    arg2, 
    arg3 %}
```

But the following, which was previously supported, will break:
```
{% shortcode arg1,,arg2,,arg3 %}
{% shortcode arg1 , arg2 , arg3 %}
{% shortcode arg1 ,arg2 ,,arg3 %}
```

There are also some changes to how shortcodes parse numbers and dot-notation variable names. The previous `moo` parser would accept numbers like `10...78` and variable names like `..foobar.baz`. These were passed to the Liquid engine, which would throw an error for invalid Liquid. The `liquid-args` parser rejects these before passing to the engine. This shouldn't cause breaking changes but could affect error handling in some cases.

Basically the same as #1269, but updated to the latest version of 11ty.


mphstudios:
@zachleat has there been any thought about merging this change or perhaps additional work to make the args lexer a configurable template engine option?

zachleat:
Closing this one per the comment left over at #1263. https://github.com/11ty/eleventy/issues/1263#issuecomment-1340145119

Again, I’m sorry this didn’t make it!

PR #1749, opened 2021-04-27
Fix for nested nunjucks async shortcodes

### LINKED ISSUE #1053: Nunjucks: async shortcode inside paired shortcode
Hi Community!

**Describe the bug**
I want to use an async shortcode inside a paired shortcode. Unfortunately there is no content from the async shortcode (child) if it is inside the paired shortcode (parent). Standalone the async shortcode works. 

**To Reproduce**
I abstracted my code to this:

**Shortcodes (inside template file)**
```
{% parentPairedShortcode %}
  {% asyncChildShortcode %}
{% endparentPairedShortcode %}
```

**Config (.eleventy.js)**
```javascript
async function getSomething() {
  const promise = new Promise((resolve, reject) => {
    setTimeout(() => resolve("done!"), 1000)
  });
  const result = await promise;
  return result;
}

eleventyConfig.addNunjucksAsyncShortcode("asyncChildShortcode", async function() {
  const something = await getSomething();
  return something;
});

eleventyConfig.addPairedNunjucksAsyncShortcode("parentPairedShortcode", async function(content) {
  return content;
});
```

**Expected behavior**
Get the content of the async shortcode (child) paired shortcode back. I tried that with liquid shortcodes and it works. Did I do something wrong? Or did I forget something?

**Environment:**
- macOS 10.15.4 
- Eleventy v0.10.0

Please let me know if more information is needed. Thank you in advance.

CodeFoodPixels:
Nested async shortcodes would fail to correctly output content from the inner shortcode as eleventy wasn't waiting for the content to have processed before passing it to the parent.

The `body()` call takes a callback, which gets called when the passed content of a paired shortcode has finished rendering, both on async and sync templates.

Fixes #1053

### DIFF
--- a/src/Engines/Nunjucks.js
+++ b/src/Engines/Nunjucks.js
@@ -295,8 +295,8 @@ class Nunjucks extends TemplateEngine {
         let body = args.pop();
         let [context, ...argArray] = args;
 
-        body(function (err, bodyContent) {
-          if (err) {
+        body(function (e, bodyContent) {
+          if (e) {
             resolve(
               new EleventyShortcodeError(
                 `Error with Nunjucks paired shortcode \`${shortcodeName}\`${EleventyErrorUtil.convertErrorToString(
--- a/src/Engines/Nunjucks.js
+++ b/src/Engines/Nunjucks.js
@@ -287,57 +287,68 @@ class Nunjucks extends TemplateEngine {
         var body = parser.parseUntilBlocks("end" + shortcodeName);
         parser.advanceAfterBlockEnd();
 
-        if (isAsync) {
-          return new nodes.CallExtensionAsync(this, "run", args, [body]);
-        }
-        return new nodes.CallExtension(this, "run", args, [body]);
+        return new nodes.CallExtensionAsync(this, "run", args, [body]);
       };
 
       this.run = function (...args) {
-        let resolve;
-        if (isAsync) {
-          resolve = args.pop();
-        }
+        let resolve = args.pop();
         let body = args.pop();
         let [context, ...argArray] = args;
 
-        if (isAsync) {
-          shortcodeFn
-            .call(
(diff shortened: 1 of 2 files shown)


src/Engines/Nunjucks.js
```diff
@@ -287,57 +287,68 @@ class Nunjucks extends TemplateEngine {
         var body = parser.parseUntilBlocks("end" + shortcodeName);
         parser.advanceAfterBlockEnd();
 
-        if (isAsync) {
```
zachleat, line 290: If I’m reading this right, this makes all Nunjucks paired shortcodes into Async shortcodes? Is this slower?

CodeFoodPixels:
Thinking about it, this is probably an issue with Nunjucks itself

CodeFoodPixels:
I've created a failing test and submitted it to Nunjucks: https://github.com/mozilla/nunjucks/issues/1363

zachleat:
Maybe related https://github.com/11ty/eleventy-img/issues/110

zachleat:
I’m interested in trying to fix this in Eleventy, if possible. I don’t think https://github.com/mozilla/nunjucks/pull/1357 is going to be enough, in local testing.

zachleat:
Filed https://github.com/11ty/eleventy/issues/2108 which is now the home base issue for this PR, I think.

zachleat:
This is a beautiful PR and I’m so sorry I slept on it for too long.

It doesn’t quite complete #2108 but it takes us a step in the correct direction.

As stated previously there may be _some_ (as of now untested) performance drawback to making everything async but we need to err on the side of correctness over performance here.

Shipping.

zachleat:
This will ship with Beta 9

zachleat:
For future me, note that the fix here was included in 1.0 stable but the async-everything approach was reverted in #2140.

PR #1764, opened 2021-05-07
TemplateConfig.js: make err.message optional

milahu:
support `throw 'whatever';` in `.eleventy.js`

otherwise

```
    TypeError: Cannot read property 'includes' of undefined
        at TemplateConfig.mergeConfig (TemplateConfig.js)
```
in
```js
err.message.includes("Cannot find module")
```


zachleat:
Ah, sorry JS optional chaining was added in Node 14 and right now we support Node 12!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

milahu:
fixed

zachleat:
Thank you!

PR #1871, opened 2021-07-01
Allow template engine overrides with addExtension, with access to the "default" engine

### LINKED ISSUE #1870: Allow template engine overrides with `addExtension`, with access to the "default" engine

**First problem: easily overriding the default markdown engine**

I noticed the ability to override template engines with `addExtension` was removed in the current v1.0 build (as of July 1st 2021). This was a more flexible solution compared to using `setLibrary` [as recommended here](https://www.11ty.dev/docs/languages/markdown/), since you could reach for any markdown library you wanted instead of being constrained to [markdown-it](https://github.com/markdown-it/markdown-it).

```js
const marked = require("marked");

...
eleventyConfig.addExtension('md', {
  read: true,
  compile: function(str) {
    return function() {
      return `<woah-thats-fancy>${marked(str)}</woah-thats-fancy>`
    }
  }
```

It could also allow you to change markdown libraries based on layout [as requested here](https://github.com/11ty/eleventy/issues/1626)!

**Second, more niche problem 🙃: modifying HTML output _before_ templates are combined into layouts**
When working on my 11ty site, I wanted to slightly modify the compiling + rendering of the default 11ty templates (Liquid and Markdown namely). I planned to add a wrapper element around each of these like so:

_Input_
```md
# Look ma, I'm a markdown file!
```

_Output_
```html
<fancy-web-component>
  <h1>Look ma, I'm a markdown file!</h1>
</fancy-web-component>
```

[HTML transforms](https://www.11ty.dev/docs/config/#transforms) seemed to fit the bill since I could access `this.inputPath` to figure out the file extension. There's a problem here though: HTML transforms operate on the output HTML file _after_ all layouts have been chained together. So this output would be impossible to automate:

_Input_
```md
---
layout: fancy.html
---

# Look ma, I'm inside a layout!
```

_Output_
```html
<layout>
  <!--notice fancy-web-component just wraps the markdown file, instead of the entire page-->
  <fancy-web-component>
    <h1>Look ma, I'm a markdown file!</h1>
  </fancy-web-component>
</layout>
```

To resolve the first problem, I'd expect this `addExtension` call to just work ™️
```js
const marked = require("marked");

...
eleventyConfig.addExtension('md', {
  read: true,
  compile: function(str) {
    return function() {
      return `<woah-thats-fancy>${marked(str)}</woah-thats-fancy>`
    }
  }
})
```

For the second problem, I would want to ability to modify the HTML output of 11ty's "default" compilers from within my `addExtension` call. I could access 11ty's markdown compiler like so:

```js
eleventyConfig.addExtension('md', {
  read: true,
  compile: function(str, inputPath) {
    return async function(data) {
      // first, call 11ty's default renderer for markdown
      const html = await this.defaultRenderer(data)
      // then, apply my custom wrapper
      return `<fancy-web-component>${html}</fancy-web-component>`
    }
  }
})
```

- HTML transforms, which are limited by layout chaining as described previously
- `setLibrary`, which limits me to the `markdown-it` library for markdown

**Additional context**
The second problem is blocking for an 11ty library I'm building called [Slinkity](https://slinkity.dev). This library can "hydrate" HTML with JavaScript on the client for all sorts of dynamic interactions, page transitions, etc.

However, my approach relies on web component wrappers to know which pieces to hydrate. So I would need a piece like this to process each file before it's flattened into a layout chain:

```js
const html = await defaultRenderer(data)
return `<fancy-web-component>${html}</fancy-web-component>`
```

Holben888:
Resolves #1870 
☝️ Full description of this change over there!

### DIFF
--- a/src/Engines/Custom.js
+++ b/src/Engines/Custom.js
@@ -76,9 +76,9 @@ class CustomEngine extends TemplateEngine {
   async compile(str, inputPath, ...args) {
     await this._runningInit();
 
-    let defaultCompiler;
+    let defaultRenderer;
     if (this._defaultEngine) {
-      defaultCompiler = async (data) => {
+      defaultRenderer = async (data) => {
         const render = await this._defaultEngine.compile(
           str,
           inputPath,
@@ -89,16 +89,17 @@ class CustomEngine extends TemplateEngine {
     }
 
     // Fall back to default compiler if the user does not provide their own
-    if (!this.entry.compile && defaultCompiler) {
-      return defaultCompiler;
+    if (!this.entry.compile && defaultRenderer) {
+      return defaultRenderer;
     }
 
     // TODO generalize this (look at JavaScript.js)
-    return this.entry.compile
-      // give the user access to this engine's default compiler, if any
-      .bind({ config: this.config, defaultCompiler })(str, inputPath)
-      // bind again for access inside function (data) {...}
-      .bind({ defaultCompiler });
+    return (
+      this.entry.compile
+        .bind({ config: this.config })(str, inputPath)
+        // give the user access to this engine's default renderer, if any
+        .bind({ defaultRenderer })
+    );
   }
 
   get defaultTemplateFileExtension() {
--- a/src/Engines/Custom.js
+++ b/src/Engines/Custom.js
@@ -10,6 +10,8 @@ class CustomEngine extends TemplateEngine {
       "init" in this.entry && typeof this.entry.init === "function";
     this.initStarted = false;
     this.initFinished = false;
+
+    this._defaultEngine = undefined;
   }
 
   getExtensionMapEntry() {
@@ -24,6 +26,10 @@ class CustomEngine extends TemplateEngine {
     throw Error(
       `Could not find a custom extension for ${this.name}. Did you add it to your config file?`
     );
+  }
+
+  setDefaultEngine(defaultEngine) {
+    this._defaultEngine = defaultEngine;
   }
 
   needsToReadFileContents() {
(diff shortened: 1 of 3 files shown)


src/TemplateEngineManager.js
```diff
@@ -56,36 +83,32 @@ class TemplateEngineManager {
...
+    let cls = this.getEngineClassByExtension(name);
 
     let instance = new cls(name, includesDir, this.config);
     instance.extensionMap = extensionMap;
     instance.engineManager = this;
 
+    // If the user providers a "Custom" engine using addExtension,
+    // But that engine's instance is *not* custom,
+    // The user must be overriding an existing engine
+    // i.e. addExtension('md', function overrideBehavior() {...})
```
Holben888, line 95:
```suggestion
    // i.e. addExtension('md', { ...overrideBehavior })
```

src/TemplateEngineManager.js
```diff
@@ -45,6 +45,33 @@ class TemplateEngineManager {
     return !!this.getClassNameFromTemplateKey(name);
   }
 
+  getEngineClassByExtension(extension) {
```
Holben888, line 48: Move this to a helper function for readability. Open to inlining this again!

zachleat (commented):
Love the tidyness of this PR! Very easily reviewable. Just a few very small points!

src/Engines/Custom.js
```diff
@@ -67,11 +73,31 @@ class CustomEngine extends TemplateEngine {
     }
   }
 
-  async compile(str, inputPath) {
+  async compile(str, inputPath, ...args) {
```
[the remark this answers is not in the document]
Holben888: @zachleat Looks like this is necessary actually! The issue is with the `defaultRenderer` that's getting defined a few lines down.

For instance, the Markdown engine takes in 4 parameters for `compile`:

```js
compile(str, inputPath, preTemplateEngine, bypassMarkdown)
```

So if your Custom engine is overriding markdown, it'll need to pass these extra args (`preTemplateEngine, bypassMarkdown`) to the markdown engine. `...args` was just a quick way to grab those extra properties. Hope that makes sense!

zachleat:
This will ship with Canary 42, thanks!

PR #1876, opened 2021-07-04
Deduplicate tags when using deepDataMerge

### LINKED ISSUE #1810: Deduplicate tags after merging
When I'm generating a collection on my site, I sometimes use data merging to provide tags common to the series and then tag individual posts further. If I happen to forget that I already have a tag for a given collection, I end up with duplicate tags. This isn't terrible, but it can easily mean that I have duplicates in the JSON Feed I generate.

Tags (and indeed data in general) should be deduplicated; at a minimum it should be *possible* to deduplicate.

- Just keep manually making sure to avoid duplicates
- Deduplicate manually anywhere the collection is used

**Additional context**
N/A

jakemulley:
Resolves #1810.

As described in #1810, tags can be duplicated through a data deep merge if they appear in more than one place.

I can't think of a reason why anyone needs a tag to appear more than once within the data cascade, so this uses `new Set` to deduplicate the output.

Hopefully this is OK!

### DIFF
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -586,6 +586,9 @@ class TemplateData {
       } else if (data.tags === null) {
         data.tags = [];
       }
+
+      // Deduplicate tags
+      data.tags = [...new Set(data.tags)];
     }
 
     return data;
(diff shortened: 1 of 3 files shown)


zachleat:
Lovely PR, thank you!

paulrobertlloyd:
@zachleat Just a note that when I upgraded to `canary.39`, I got an error building my site. This is because I was using `tags: false` on some of my pages, instead of what I should have been using, `tags: []`. Do we need to guard against cases where `data.tags` isn’t an array?

pdehaan:
Interesting! Looks like `permalink: false` "worked" in 0.12.1, but throws an error in 1.0.0-canary.39.

Issue #1891, opened 2021-07-13
Remove fs-extra dependency


zachleat:
With the `fs.mkdir` recursive option and the promises API on all Node versions we support I think we can start phasing this one out.

Snapstromegon:
Related to this issue:
Maybe "parse-filepath" can be removed as a dependency too.
According to my quick search in the code it seems like ```path.parse``` should be a working replacement.

zachleat:
Merged, thanks @Snapstromegon !

PR #1896, opened 2021-07-20
Remove fs-extra dependency (#1891)

Snapstromegon:
@zachleat proposed in #1891 removing fs-extra since the build in FS module now provides a promise based option and supports recursive options.

This Commit only uses the ```require("fs")``` instead of ```require("fs").promises``` or ```require("fs/promises")```, since the last one is not support in Node Version <14 and the sync API is used.

Except for Template.js no file logic needed to be changed to remove fs-extra.

According to my run of "npm test" test results are not changed by this commit.

This commit only replaces fs-extra with fs and does not touch existing logic when it's not needed. Maybe some optimizations like in 5ca1aa755dfb0fac76a86aba360c0cd7a78e780d are possible.

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -1,4 +1,5 @@
-const fs = require("fs-extra");
+const fs = require("fs");
+const path = require("path");
 const parsePath = require("parse-filepath");
 const normalize = require("normalize-path");
 const isPlainObject = require("lodash/isPlainObject");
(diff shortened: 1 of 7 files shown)


src/Template.js
```diff
@@ -751,11 +752,11 @@ class Template extends TemplateContent {
     } else {
       let templateBenchmark = bench.get("Template Write");
       templateBenchmark.before();
-      return fs.outputFile(outputPath, finalContent).then(() => {
-        templateBenchmark.after();
-        this.writeCount++;
-        debug(`${outputPath} ${lang.finished}.`);
-      });
+      await fs.promises.mkdir(path.parse(outputPath).dir, { recursive: true });
```
zachleat, line 755: This is not _quite_ the same but we’ll fix it in post!!

zachleat:
Thank you!

Issue #1903, opened 2021-07-26
Remove `parse-filepath` dependency


zachleat:
Replace with `path.parse` as noted by @Snapstromegon at https://github.com/11ty/eleventy/issues/1891#issuecomment-882458675

Make sure we double check the behavior of `path.parse` when you feed it a directory in addition to the standard use case of a file path.

Snapstromegon:
According to my short test ```path.parse``` and ```parse-filepath``` behave exactly the same:

```js
const parseFilepath = require("parse-filepath");
const nativePath = require("path");
const assert = require("assert");

// Simple List of Testcases, easy to expand
const testcases = [
  "/",
  ".",
  "asd/asdf/",
  "asd/asdf/test.txt",
  "./",
  "./test.js",
  "./test",
  "./..test",
  "./.test",
  "./.asd.test",
  "./testa",
];

for (const dir of testcases) {
  // Remove the items from the parse-filepath result, which are not returned by the native path module.
  // This should not be a problem for 11ty, since at the moment those aren't used and native modules provide alternatives.
  const {
    absolute,
    basename,
    dirname,
    extname,
    isAbsolute,
    path,
    stem,
    ...parsedByParseFilepath
  } = parseFilepath(dir);
  const parsedByPath = nativePath.parse(dir);
  assert.deepStrictEqual(parsedByPath, parsedByParseFilepath);
}
```

(Tested on the parse-filepath version from current master.)

zachleat:
Awesome!

zachleat:
Fixed by #1904

PR #1904, opened 2021-07-26
Remove parse-filepath dependency

Snapstromegon:
As described in #1903 this commit removes the parse-filepath dependency from 11ty in favor of the native path.parse.

Some files needed renaming of some variables which were named "path" and shadowed the path module. I hope I selected reasonable replacements.

In TemplatePath.js I also changed the documentation links to point to the nodejs api doc instead of the npm package.

According to a local test run it does not change test results.


zachleat:
Awesome, thank you!!

PR #1909, opened 2021-07-28
docs: document capitalize utility

Ryuno-Ki:
This function is mainly used as filter. It assumes standard Unicode
case mappings. That is, it can have surprising for certain locales
such as Turkey:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase

I was wondering, whether a lesson could be learned by Lodash,
which defines several helper functions to transform into different
casings. This function here could then defer to the respective
transformation.

### DIFF
--- a/src/Util/Capitalize.js
+++ b/src/Util/Capitalize.js
@@ -1,3 +1,11 @@
+/**
+ * Capitalize a given string.
+ *
+ * @param {string} str - The string to capitalize.
+ * @param {Object} options - Additional options.
+ * @param {boolean} options.lowercaseRestOfWord - The only recognised option. Transform the rest of the input?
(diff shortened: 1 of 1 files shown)


zachleat:
I don’t think that utility is even being used in a way that makes sense. Might make more sense to refactor it out

Ryuno-Ki:
I can see it's used, but without the option. Thus, it can be simplified. Shall I refactor?

zachleat:
Ah sorry I just went ahead and got rid of it—thanks for the prompt!

Ryuno-Ki:
No worries. I'm going on the hunt for another function then :-)

PR #1912, opened 2021-07-29
Fix #1242 Dot Pathing in data file paths

### LINKED ISSUE #1242: Dots in Global Data Files are used for dot-pathing instead of preserved in key name
**Describe the bug**

**To Reproduce**
In an eleventy project:

1. Create a file at `_data/sites/11ty.dev.json` that just contains:
    ```json
    {}
    ```
2. Create a file at `index.liquid` that just contains:
    ```liquid
    {{sites | log}}
    ```
3. Run `npx @11ty/eleventy`
4. See console output:
    ```js
    { '11ty': { dev: {} } }
    ```

**Expected behavior**
I don't know if this is actually expected behaviour, but I would have assumed it would output 

```js
{ '11ty.dev': {} }
```

The 11ty docs seem to operate under this assumption, because [sites with dots in the filename don't show up on the leaderboard due to the unexpected dot-path behaviour](https://github.com/11ty/11ty-website/issues/639).

**Environment:**

- OS and Version: macOS 10.15.5
- Eleventy Version: 0.11.0

**Additional context**
11ty/11ty-website#639

I have a commit that could be turned in to a PR if this is actually a bug.

Snapstromegon:
Dots in the global datafile path leads to data beeing incorrectly nested.
This commits changes the behavior so a datafile a.b.json will create an entry with data["a.b"] instead of data.a.b.
The old behavior was a bug according to 11ty#1242.

The implementation makes use of lodashs set and get functions which also take object paths in array form instead of string. That way we can implement this safely.


src/TemplateData.js
```diff
@@ -265,21 +265,27 @@ class TemplateData {
     let dataFileConflicts = {};
 
     for (let j = 0, k = files.length; j < k; j++) {
-      let objectPathTarget = await this.getObjectPathForDataFile(files[j]);
       let data = await this.getDataValue(files[j], rawImports);
+      let objectPathTarget = this.getObjectPathForDataFile(files[j]);
```
Snapstromegon, line 269: I removed the await here, since getObjectPathForDataFile is returning a String[] anyways, so there is no Promise to await.

Snapstromegon:
I think this PR introduces a major version bump change, because it breaks compatibility to existing projects that might use the bug this fixes as a feature. Because of this I would recommend to either make the "bug" in #1242 an official feature, or to include this in 1.0.

PR #1914, opened 2021-07-30
docs: document ConsoleLogger class

Ryuno-Ki:
VS Code is warning me (via [CodeMetrics extension](https://marketplace.visualstudio.com/items?itemName=kisstkondoros.vscode-codemetrics)), that this class has become quite complex.

As far as I can tell, it's the only logger implementation you have in this repository. That being said, it is possible to overwrite the logger. Given that, I suggest, that you consider defining a BaseLogger class that developers can extend with their own custom logic (for example, use pino for logging).
This way, you can use the class-as-interface pattern while reducing the overhead in this class.

I was a bit unsure, which properties and methods you consider public and which private/protected. I assume, that a leading underscore indicates a „private” property.
Since the variables are pretty self-explaining, I omitted the description blocks in most of them.


src/Util/ConsoleLogger.js
```diff
@@ -2,10 +2,16 @@ const chalk = require("chalk");
 const debug = require("debug")("Eleventy:Logger");
 const Readable = require("stream").Readable;
 const split = require("split");
+/** @typedef {import('stream').Readble Readable} */
```
pdehaan, line 5: I'm a bit confused here. Is `Readble` a typo, or am I missing something?

Ryuno-Ki: Obviously a typo m(

Ryuno-Ki:
@zachleat @pdehaan What's blocking this PR from moving forward?

pdehaan:
Looks good to me, but I don't have merge permissions on this repo so I can't land it.

zachleat:
Will ship with 1.0.1, thanks!

PR #2040, opened 2021-10-20
should not render page entry if file should not be written

magua-io:
This commit helps me shorten the build time for the repo that contains 200+ markdown files. The --incremental build time was shorten from ~90 seconds to ~3 seconds when I only made some changes in one Markdown file.

Please review and decide whether the commit is valid.

Thanks!

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -738,7 +738,7 @@ class Template extends TemplateContent {
     return pages;
   }
 
-  async _write(outputPath, finalContent) {
+  async _write(outputPath, mapEntry, page) {
     let shouldWriteFile = true;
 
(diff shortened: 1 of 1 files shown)


zachleat:
Ah, my first reaction here is that this won’t accommodate collection use right? This is why incremental needs to know where collection `templateContent` is used.

zachleat:
Gotta close this one for now unfortunately but we will circle back on this once better incremental support lands. 

This is logged as “Smarter dependency map between templates” on https://www.11ty.dev/docs/usage/incremental/#templates-2

zachleat:
This improvement will ship with 2.0.0-canary.21

Issue #2033, opened 2021-10-17
Explain how to use JSDoc with 11ty config file


wdsrocha:
Hi 👋

Some time ago @Ryuno-Ki added JSDoc support in the following PR: https://github.com/11ty/eleventy/pull/720

My question is: how to use it? It might be silly and something that anyone with basic knowledge in JSDoc knows how to do, but I have no clue on how to make it work. I've tried:

- Searching for someone who did it already, but I couldn't find it even in Ryuno-Ki public repositories
- Searching docs/tutorials on the internet. All that I found are the Ryuno PR and https://github.com/11ty/eleventy/issues/1459
- Trial and error, but failed in all attempts and I'm not sure if it's because of JSDoc, my IDE (VSCode), Eleventy or something else

Here is my last attempt

```js
/** @type {import('@11ty/eleventy').Eleventy} */
module.exports = function (config) { 
  return {
    dir: {
      input: "src",
      output: "build",
    },
  };
};
```

If this isn't something straightforward and there is anything that I can do to help, please tell me. Thanks!

paulshryock:
@wdsrocha, based on [this type definition](https://github.com/11ty/eleventy/blob/master/src/TemplateConfig.js#L18), does this work?
```javascript
/** @type {import('@11ty/eleventy').TemplateConfig~TemplateConfig~config} */
module.exports = function (config) {
  // do something
}
```

wdsrocha:
It does not, @paulshryock. But I've noticed that the JSDoc is missing in the build files (that is `node_modules/@11ty`) - could that be the problem?

wdsrocha:
But I just found out that this _seems_ to work

```js
const UserConfig = require("@11ty/eleventy/src/UserConfig");

/** @param {UserConfig} config */
module.exports = function (config) {
  config.addPassthroughCopy("./src/*.css"); // this was autocompleted

  return {
    dir: {
      input: "src",
      output: "dist",
    },
  };
};
```

I'm just not sure if UserConfig is a subset of what this configuration parameter is supposed to be

wes566:
I opened #2091 to hopefully make this easier for consumers.  With that you can import from the package without knowing the internal paths of things:

```js
/** @param {import("@11ty/eleventy").UserConfig} eleventyConfig */
module.exports = function (eleventyConfig) { ... }
```

FunctionDJ:
This seems like the leanest method right now, which is fine once you know about it:
```js
[...]

PR #720, opened 2019-10-03
Add JSDoc Comments for Config and TemplateConfig. Adjust references

Ryuno-Ki:
Hi @zachleat,

today is bank holiday in my country so I took some time to ad moar JSDoc strings :-)

If you are fine with, I'd submit those additions in small chunks of like 100 lines of code each (to ease the review).

Cheers

Ryuno-Ki

PR #2091, opened 2021-11-16
feat: declare "types" in package.json to make consuming types easier

wes-goulet:
Hi 👋  - long time listener, first time caller!

I like to type my javascript files using JSDoc comments.  Thanks to an [excellent contribution](https://github.com/11ty/eleventy/pull/720) this is possible for the `.eleventy.js` config file 🎉 .  However, it can be more discoverable/easier with a small tweak 😄 ... that's where this here PR comes.

As discussed in #2033 the the way for 11ty consumers to get types is to reach into the 11ty package internals and find the type you want.  That's less than ideal for consumers and a bit fragile (if the file ever moves with a refactor or whatnot).

So this PR does 2 things:
1️⃣  - Makes an index.d.ts declaration file that can export types
2️⃣  - updates the package.json "types" property with the path to that type declaration file

So now a consuming project can import directly from the eleventy package without reaching into it to a specific file path:

```js
/** @param {import("@11ty/eleventy").UserConfig} eleventyConfig */
module.exports = function (eleventyConfig) { ... }
```

This PR just starts with adding the `UserConfig` to get it up and running, but additional types can be exported from `index.d.ts` over time.

> As a personal preference thing, I like to use typedef's to pull in all my dependency types at the top of the file, like so 👇 

```js
/**
 * @typedef { import("@11ty/eleventy").UserConfig } UserConfig
 */

/** @param {UserConfig} eleventyConfig */
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,3 @@
+import UserConfig from "./UserConfig";
+
+export { UserConfig };
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "access": "public"
   },
   "main": "src/Eleventy.js",
+  "types": "src/index.d.ts",
   "bin": {
     "eleventy": "./cmd.js"
   },


pdehaan:
Related discussion from earlier today: https://github.com/11ty/eleventy/discussions/2089

src/index.d.ts
```diff
@@ -0,0 +1,3 @@
+import UserConfig from "./UserConfig";
+
+export { UserConfig };
```
pdehaan, line 3: Per #2089, do we want to reexport the `defaultConfig` as well?
```js
 * @typedef {import("@11ty/eleventy/src/UserConfig")} EleventyConfig
 * @typedef {ReturnType<import("@11ty/eleventy/src/defaultConfig")>} EleventyReturnValue
```

wes566: That's a good question, I don't think so yet.  I was thinking of keeping the scope of this PR down to just setting up the initial type declaration file with the `UserConfig` (which is a well-defined class, making it a good initial type to export).  I was thinking exporting additional types can come after this, wanted to keep this one small, easier for @zachleat to review (and hopefully merge 😄 ).

BTW, getting that `EleventyReturnValue` brings up a few questions... should it return a `Partial` of that return type (since user config doesn't require all those types, they are all optional - they get merged with the defaultConfig)?  Should we make an actual explicit interface for `EleventyReturnValue` (the `ReturnType<typeof defaultConfig>` method of inferring types is missing some properties, like `dir.layouts`)?  Does this project want to have explicitly defined types/interfaces or just let typescript infer types based on the JS?  I personally like the "contract" of having explicitly defined types, but I'm just some random dude so I didn't want to include such decisions in this PR 😀

FunctionDJ:
This would still be really cool to have because i tried JSDoc `import("11ty/eleventy").` before even looking up the "official" method (see issue below) :eyes:
https://github.com/11ty/eleventy/issues/2033#issuecomment-947373901

zachleat:
It’s happening! It’s all happening! 2.0.0-canary.19

paulshryock:
@zachleat was this reverted or removed? As of `3.0.0-alpha.10` it seems like `types` is no longer set in `package.json`.

chriskirknielsen:
I've noticed the same as Paul. I modified my local files to restore the feature, seems to be working as expected. It might have been an incorrect diff resolution between `"types": "src/index.d.ts"` and `"type": "module"`? (though the `index.d.ts` file itself also disappeared)

If it was accidental, I've raised a PR to restore this: https://github.com/11ty/eleventy/pull/3291

zachleat:
#3291 was merged and is shipping in `3.0.0-alpha.11`

PR #2112, opened 2021-11-20
Global Data Files should be more specific in data cascade than Custom Global Data

vdse:


zachleat:
Howdy—what’s the reasoning here?

I think I can be persuaded but not with `No description provided.` 😅

vdse:
Documentation define this order for data cascade:
1. Computed Data
2. Front Matter Data in a Template
3. Front Matter Data in Layouts (only in 0.x)
4. Template Data Files
5. Directory Data Files (and ascending Parent Directories)
6. Front Matter Data in Layouts (moved in 1.0)
7. Global Data Files
8. Custom Global Data

If I add ```cascade``` object in ```.elevenyt.js``` like this:
```
  eleventyConfig.addGlobalData("cascade", {priority: 8});
}
```
And then print on page
```cascade.priority is 8```

After create file ```_data/cascade.json:```
```
{
  priority: 7
}
```
On page I still have
```cascade.priority is 8```

Regarding documentation I expect that ```cascade.priority must be 7```

If it's would be helpful I can create small working example that demonstrate this.

zachleat:
A ha! Yes, sorry that is the intended behavior—the docs are wrong. Configuration API Global Data should override Global Data Files in the data cascade.

I just pushed a fix to the docs to make them accurate—sorry about that!

PR #2188, opened 2022-01-25
feat: pass file path to custom data parser

DamianOsipiuk:
This change is to pass the currently processed file path to the custom data parser.

Reason to make this change is to be able to process the file along with it's dependencies that are imported with relative paths.
To do that parser needs to know what is the entry file path to be able to resolve relative imports.

Custom template parser provides file path already, but custom data parser provides only text content of the file.

### DIFF
--- a/src/TemplateData.js
+++ b/src/TemplateData.js
@@ -423,7 +423,7 @@ class TemplateData {
 
     if (ignoreProcessing || engineName === false) {
       try {
-        return parser(rawInput);
+        return parser(rawInput, path);
       } catch (e) {
         throw new TemplateDataParseError(
           `Having trouble parsing data file ${path}`,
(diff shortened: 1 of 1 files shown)


zachleat:
This sounds fine to me but there’s a small code path not handled, specifically when JSON.parse is fed into that function via https://github.com/11ty/eleventy/blob/aba3a57a74d0f9e481e3608e9d17c0a2ace5dc3b/src/TemplateData.js#L503

When a second argument is passed to JSON.parse it’s treated as a `reviver` function, which is not what we want here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#parameters

I think just wrap that JSON.parse in another function and we should be good

zachleat (changes requested):
Per https://github.com/11ty/eleventy/pull/2188#issuecomment-1045268230

DamianOsipiuk:
@zachleat Are there any more changes that are required?

zachleat:
Shipping with 2.0.0-canary.19

zachleat:
Docs building to https://www.11ty.dev/docs/data-custom/

Issue #2214, opened 2022-02-10
Consider faster alternatives to lodash/isPlainObject


jeffposnick:
**Is your feature request related to a problem? Please describe.**
I'm taking a look at the build performance of https://github.com/GoogleChrome/web.dev, which is a fairly large (~3300 files) 11ty site.

**Describe the solution you'd like**
A detailed description of one of my initial findings is at https://github.com/GoogleChrome/web.dev/issues/7339; the takeaway is that https://github.com/11ty/eleventy/blob/08fe54e03da025cda24c4b25b5b8a8d0007b4689/src/Util/Merge.js#L1 ends up being a very big performance hotspot for our build. Using [`module-alias`](https://github.com/ilearnio/module-alias) to swap out [`lodash/isPlainObject`](https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/isPlainObject.js) for the implementation from [`typechecker`](https://github.com/bevry/typechecker), without making any other changes, takes our build time down from **55 seconds** to **40 seconds**.

It would be great if someone from the core 11ty team could try making this change as well and see if they experience similar performance improvements on large sites, without any unintended changes.  I'm assuming the 11ty test suite would pick up any compatibility issues.

**Describe alternatives you've considered**
I can't guarantee whether the implementation in `typechecker` is 100% compatible with `lodash/isPlainObject`, or whether `typechecker`'s implementation is faster than a few other implementations that I found on `npm`.

**Additional context**
As mentioned, there are details in https://github.com/GoogleChrome/web.dev/issues/7339 of the methodology for this investigation and steps to reproduce what I tried.

jeffposnick:
It's also worth noting that even with the performance boost we get from swapping out `isPlainObject`, https://github.com/11ty/eleventy/blob/master/src/Util/Merge.js accounts for a large chunk of our build's execution time.

We are pre-v1.0.0 but [explicitly opted-in](https://github.com/GoogleChrome/web.dev/blob/main/.eleventy.js#L242) to [data deep merge](https://www.11ty.dev/docs/data-deep-merge/). I believe that it's on by default for everyone post-v1.0.0, so maybe it's worth taking a pass at optimizing the merge algorithm? I could imagine other sites that didn't have data deep merge turned on seeing a slowdown in v1.0.0 if they suddenly start hitting this code path.

jeffposnick:
[...]

PR #2219, opened 2022-02-15
Swap to use internal `isPlainObject` utility

zachleat:
In service of: https://github.com/11ty/eleventy/issues/2214


src/Util/Merge.js
```diff
@@ -1,4 +1,4 @@
-const isPlainObject = require("lodash/isPlainObject");
+const isPlainObject = require("../Util/IsPlainObject");
```
jeffposnick, line 1: Nit: you're already in `Util/` so this could be `require("./IsPlainObject");

jeffposnick:
👍

Issue #2248, opened 2022-02-26
Allow Custom templates to alias to an existing template engine


pspeter3:
**Is your feature request related to a problem? Please describe.**
Allow users to run eleventy through Babel, TypeScript, ESBuild, or others without registering a custom template engine.

**Describe the solution you'd like**
Add `11ty.js`, `11ty.ts`, and `11ty.tsx` to the TemplateEngineManager. https://github.com/11ty/eleventy/blob/master/src/TemplateEngineManager.js#L26.

**Describe alternatives you've considered**
Allow users to add entries to the extension map. This would also work just fine.

pspeter3:
This idea from @zachleat seems like a great alternative as well https://twitter.com/zachleat/status/1473036576468918279.

zachleat:
I am so sorry for the super-late reply here.

This is _almost_ possible today in Eleventy 1.x using current APIs:

```js
eleventyConfig.addExtension("11ty.tsx", {
	key: "11ty.js",
	read: false,
});
```

I don’t consider needing `read: false` to be great there so I did push a fix that will inherit this from the upstream alias (shipping in 2.0.0-canary.19). That said, I tested the above on your excellent `gist` and it does write the file and the HTML looks okay.

```js
> build
> node --require esbuild-register node_modules/.bin/eleventy

[11ty] Writing _site/index.html from ./index.11ty.tsx
[11ty] Wrote 1 file in 0.03 seconds (v2.0.0-canary.18)
```

**There is one big problem, it isn’t fetching the data out of the file for the data cascade.**

I’m going to continue this one tomorrow, I think configuration-enabled aliasing will be useful so I don’t have to be the arbiter of these moving forward.

---
Aside, I do think it’s worthwhile to allow multiple aliases in a single command, so I added that to 2.0.0-canary.19:

```js
// v2.0.0-canary.19 or newer
eleventyConfig.addExtension([ "11ty.jsx", "11ty.ts", "11ty.tsx" ], {
	key: "11ty.js"
});
```

pspeter3:
Thanks!

> **There is one big problem, it isn’t fetching the data out of the file for the data cascade.**

That's what I was trying to solve for with this suggestion.

zachleat:
Updated your gist with the new API: https://gist.github.com/zachleat/b274ee939759b032bc320be1a03704a2

zachleat:
Docs are building to https://www.11ty.dev/docs/languages/custom/#aliasing-an-existing-template-language

[...]

Issue #577, opened 2019-06-19
How to use TypeScript with Eleventy?


d2s:
Are there examples of how to use TypeScript with Eleventy?

Would like to be able to use TS with 11ty.

Ryuno-Ki:
Have you tried it yourself, yet?
If so, where did you get stuck?
(I'm not a fan of TypeScript, personally).

hcavalieri:
As far as I understand, the only current way to do so is through using the `--watch` flag that will transpile the code whenever there's a change; eleventy then picks up on these changes through its own `--watch` and rebuilds corresponding part of the website.

In theory, this works fine, but it presents a huge burden on our workload, as you now have to **wait for 2 processes to happen in sequence before you can see any change** propagate. I've recently gotten into the same situation when trying to use [Svelte](https://svelte.dev/) as a template engine for specific parts of my site, and would love to know if there's any specific implementation that mitigates this problem.

My current implementation is as follow:

1. Have [Rollup](https://github.com/rollup/rollup) bundle my Svelte app with a Server (SSR) and a Client (with hydration) configuration;
1. Set it to watch changes to my Svelte directory, which triggers a new Rollup build;
1. Then inside my `.11ty.js` template I import the `Server.js` bundle, run the `SvelteApp.render()` method that return an `html` string, which I then return in the Eleventy layout;
1. For client-side hydration, I also include a `<script src="svelteApp.js" defer>` that refers to the client bundle generated by Rollup;
1. This client bundle is generated inside a mid-way `dist` folder that Eleventy then picks up through `addPassthroughCopy`;
1. Finally, I run `eleventy --serve` and hope for the best.

**Problems:**

- Every little change has to go through Rollup _and_ Eleventy. This is not a problem with a small codebase like the one I'm fiddling with, but it could become one for full-blown projects;
- For some strange reason, `.11ty.js` templates don't update until I run `eleventy --serve` again, even though Eleventy does notice a change and rebuilds the website, very odd behavior. This makes testing the SSR impossible while developing, as changes to the `Server.js` don't get reflected in the browser;
- And, most importantly, the above problem would _suck_ for developing Typescript `.11ty.ts` templates, as they simply wouldn't change. Worth the shot, though!

**NOTE:** I know this is not related directly to Typescript, but as I'm dealing with transpilation much like you'd with TS, this might be of help. If needed, I can transfer the svelte discussion somewhere else 🤗

jevets:
[...]

PR #2249, opened 2022-02-26
Configure more extensions as the JavaScript engine

pspeter3:
Allow users to treat JSX, TS, and TSX files as the JavaScript engine.

https://github.com/11ty/eleventy/issues/2248

This is based on the comment from https://github.com/11ty/eleventy/issues/577#issuecomment-1037632961


pspeter3:
This is actually the wrong implementation. I just tested it locally and have a different version.

Issue #1522, opened 2020-11-21
Normalize signature of `this` in callbacks that operate on templates


zachleat:
Thinking https://www.11ty.dev/docs/config/#transforms and linters and `before` callback in Pagination (#1209).

This will be important before 1.0 release to avoid any breaking changes.

I’d like the metadata available in `this` to be consistent here. Specifically `this.page.inputPath` instead of `this.inputPath`. See https://www.11ty.dev/docs/data-eleventy-supplied/

There are likely others—I don’t have time to compile them all here right this second but we’ll need to look through and find them all.

zachleat:
* [ ] Linters
* [ ] Transforms
* [ ] Built-in Filters
* [ ] Shortcodes
* [ ] Configuration Function (probably not?)
* [ ] Plugin config functions (probably not?)

zachleat:
Related to https://github.com/11ty/eleventy/issues/338 (and https://github.com/11ty/eleventy/issues/130)

There are two styles in play here:

Collection-style:
* inputPath, fileSlug, outputPath, url, date, data (full data cascade), templateContent

Template-style:
* everything goes through `page`

---

## ACTION ITEMS

* [ ] Surface `page` in collection-style entries for more consistency (while keeping backwards compatibility). This would solve some of the issues in #338 (but not top level `data` and `content`)
* [ ] Surface `page` in Linters
* [ ] Surface `page` in Transforms
* [ ] Surface `page` in Filters
* [x] Shortcodes (this is already done—you can use this.page on Handlebars, Liquid, 11ty.js, Nunjucks)

zachleat:
After this last analysis I don’t think this is a blocker for 1.0. Moving it out of the milestone.

zachleat:
Filters/helpers shipped with #2250

zachleat:
`this.page` is available on Shortcodes, Filters, Linters, Transforms in 2.0.0-canary.19 and newer

zachleat:
`page` and `content` (aliased to `templateContent`) is available on Collections items on 2.0.0-canary.19 and newer. https://www.11ty.dev/docs/collections/#collection-item-data-structure

nhoizey:
Time to move to the 2.x branch I guess!

zachleat:
Cross linking the shortcode work that shipped with v0.11.0 #741

PR #2250, opened 2022-02-27
Standardized `this` signature for filters

LeoSchae:
Adds a more consistent `this` signature to Nunjucks and Liquid filters, as well as Handlebars helpers and JavaScript functions. Related to #1522.

The changes aim to make the this signature consistent for all filters added using `eleventyConfig.addFilter(...)` as well as the individual methods for the individual template languages. Example:
```javascript
eleventyConfig.addFilter("pagetitle", function(arg1, arg2) {
    console.log("Called 'pagetitle' from " + this.page.inputPath);
    return this.ctx.title;
})
```

## Details

The format of the this object is given by
```javascript
{
    ctx: {
        title: "...",
        content: "...",  // (only in templates)
        collections: [...],
        ...
    },
    page: {
        inputPath: "...",
        outputPath: "...",
        url: "...",
        ...
    }
}
```
It should still be discussed if this is a "good" format or if a better one could be found. This current version is a mix from the default nunjucks (which provides the ctx field) and nunjucks shortcodes (which provide the page field).

The `ctx` field contains the context (variables) of the templating engine or the [Edit: ~~JavaScript module~~ data cascade (latest commit)] for 11ty.js files.

The actual `this` object may contain more fields. For backwards compatibility the original fields provided by the templating engine are inherited.

## Potential problems

~~The ctx field in 11ty.js files does not contain the data cascade but only the module itself.~~ [Edit: Fixed with latest commit]

Almost all projects should be unaffected. Projects can break when writing to fields of the `this` object.

### DIFF
--- a/src/Engines/Handlebars.js
+++ b/src/Engines/Handlebars.js
@@ -27,22 +27,20 @@ class Handlebars extends TemplateEngine {
     this.handlebarsLib.registerHelper(name, callback);
   }
 
-  wrapHelper(callback) {
-    let jsFuncs = this.config.javascriptFunctions;
+  static wrapHelper(callback) {
     return function () {
       const newThis = {
-        ...jsFuncs,
+        ...this,
         ctx: this,
-        page: this.page,
+        // page: this.page
       };
-      // Handlebars passes an additional argument
       return callback.call(newThis, ...arguments);
     };
   }
 
   addHelpers(helpers) {
     for (let name in helpers) {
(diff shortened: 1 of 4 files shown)


LeoSchae:
Added a patch that provides the full data cascade in `this.ctx` for 11ty.js files instead of the module itself. With this every format has access to the `this.ctx.collections` field.

zachleat:
Made a few tweaks to this post-merge but it will ship with 2.0.0-canary.19! Thank you!

PR #2279, opened 2022-03-16
Alias more extensions as 11ty.js

pspeter3:
Allow more extensions to be treated at the `11ty.js` template engine. This builds on the work of #2249 and solves #2248. A user could run Eleventy via `ts-node` or `esbuild-register` and have the appropriate behavior now.

### DIFF
--- a/src/EleventyExtensionMap.js
+++ b/src/EleventyExtensionMap.js
@@ -239,6 +239,9 @@ class EleventyExtensionMap {
         liquid: "liquid",
         "11ty.js": "11ty.js",
         "11ty.cjs": "11ty.js",
+        "11ty.jsx": "11ty.js",
+        "11ty.ts": "11ty.js",
+        "11ty.tsx": "11ty.js",
       };
 
       if ("extensionMap" in this.config) {


pspeter3:
In my example, I was able to run `node --require esbuild-register node_modules/.bin/eleventy --config=.eleventy.ts` to have a configuration file and layouts that were in TypeScript.

pspeter3:
@zachleat is there a way to make this change to the extension map from the `eleventyConfig` object instead of baking it into the framework by default?

DamianOsipiuk:
@pspeter3 You could create a typescript plugin using eleventy 1.x built-in methods.
I use similar setup to add TS support in my project.

Sample config using `esbuild`:
```typescript
config.addTemplateFormats("11ty.ts");
config.addExtension("11ty.ts", {
  outputFileExtension: "js",
  compile: (_fileContent: string, filePath: string) => {
    return () => {
      return buildSync({
        entryPoints: [filePath],
        write: false,
        bundle: true,
        treeShaking: true,
        minify: options.minify,
        sourcemap: options.sourcemaps,
      }).outputFiles[0].text;
    };
  },
});
```

The same can be done for data files, but support for it is pending: https://github.com/11ty/eleventy/pull/2188

pspeter3:
Thanks @DamianOsipiuk. Does this cause the TypeScript files to be treated as normal `11ty.js` files after the transformation?

DamianOsipiuk:
No, this example is only bundling and returning JavaScript, but you could extend it to add the same features.

pspeter3:
I don't want to have to extend it though. I would ideally like to have the features of the `11ty.js` extension by default which I think this would give.

zachleat:
Can you attach a demo repo or test case showing how this would be used with sample inputs?

pspeter3:
Yup! I'll make test case next week.

pspeter3:
@zachleat I made an example repo. I had to remove the `patch-package` patch since Github would not let me push it to a Gist. It should work otherwise. You can see `index.11ty.tsx` as the example.

pspeter3:
Oops! Forgot the link to the Gist https://gist.github.com/pspeter3/56931a837cb854c55bb06024287ead95

pspeter3:
Just following up, is there something else I can do to help here?

pspeter3:
Hi! It's been a few months and I was wondering if this could get merged?

pspeter3:
Hi! It's been a few months and I'm wondering what I can do to help get this merged?

zachleat:
I am so sorry for the super-late reply here.

This is _almost_ possible today in Eleventy 1.x using current APIs:

```js
eleventyConfig.addExtension("11ty.tsx", {
	key: "11ty.js",
	read: false,
});
```

I don’t consider needing `read: false` to be great there so I did push a fix that will inherit this from the upstream alias (shipping in 2.0.0-canary.19). That said, I tested the above on your excellent `gist` and it does write the file and the HTML looks okay.

```js
> build
> node --require esbuild-register node_modules/.bin/eleventy

[11ty] Writing _site/index.html from ./index.11ty.tsx
[11ty] Wrote 1 file in 0.03 seconds (v2.0.0-canary.18)
```

**There is one big problem, it isn’t fetching the data out of the file for the data cascade.**

I’m going to continue this one tomorrow, I think configuration-enabled aliasing will be useful so I don’t have to be the arbiter of these moving forward.

---
Aside, I do think it’s worthwhile to allow multiple aliases in a single command, so I added that to 2.0.0-canary.19:

```js
// v2.0.0-canary.19 or newer
eleventyConfig.addExtension([ "11ty.jsx", "11ty.ts", "11ty.tsx" ], {
	key: "11ty.js"
});
```

zachleat:
Moving my above comment to #2248

pspeter3:
That sounds great, I'm looking forward to the aliasing solution!

PR #2306, opened 2022-04-04
Add types for Render plugin

binyamin:
I decided to write it as a separate file, so that I could use `@default` to assign default values to the `options` parameter's children.


pdehaan (approved):

binyamin:
@zachleat do you mind merging this?

zachleat:
Shipping with 2.0.0-canary.19

PR #2358, opened 2022-05-03
Fix #2350

### LINKED ISSUE #2350: Custom template fails to render in v1.0.1
**Describe the bug**
I'm using Eleventy custom templates to handle sass files with template skips via .startsWith('_'); as described by this example here: https://www.11ty.dev/docs/languages/custom/#skipping-a-template-from-inside-of-the-compile-function. However since updating to Eleventy v1.0.1 my test builds have started to fail with the following error:

```
[11ty] 1. Having trouble rendering scss template ./source/assets/styles/_callout.scss (via TemplateContentRenderError)
[11ty] 2. Having trouble compiling template ./source/assets/styles/_callout.scss (via TemplateContentCompileError)
[11ty] 3. Cannot read properties of undefined (reading 'bind') (via TypeError)
[11ty] 
[11ty] Original error stack trace: TypeError: Cannot read properties of undefined (reading 'bind')
[11ty]     at *[project path]*/node_modules/.pnpm/@11ty+eleventy@1.0.1/node_modules/@11ty/eleventy/src/Engines/Custom.js:173:35
```

**To Reproduce**

1. Follow example from https://www.11ty.dev/docs/languages/custom/ on adding sass support to Eleventy
2. Implement early returning from the `compile()` function with a `.startsWith('_')` conditional as described here: https://www.11ty.dev/docs/languages/custom/#skipping-a-template-from-inside-of-the-compile-function
3. Run build

**Expected behavior**
Given no apparent syntax errors in the compile function code, I expected that Eleventy v1.0.1 would process the scss files and output them to the output directory

**Environment:**

- Ubuntu 21.10 via kernal 5.10.102.1-microsoft-standard-WSL2
- Eleventy Version v1.0.1

**Additional context**
1. What's odd to me is that the error persists even when returning early from the compile function, meaning that something is catching on the eleventy side rather than as an issue with my implementation or with `sass`, it's puzzling to me why the error is being thrown when attempting to compile a file that should be skipped; `_callout.scss` in this case.
2. Reverting to Eleventy v1.0.0 appears to fix this issue so it's definitely a byproduct of updating
3. Creating a reduced test case with just the configuration required and a couple of .scss files doesn't appear to reproduce the issue which means that there's something specific to my project that was working on v1.0.0 and now on v1.0.1 it's not. Happy to help triage the issue if Zach or a contributor has any ideas on what to try next.

kentaroi:
Currently, it fails when a custom template extension has an async compile function and it returns undefined to ignore, because there is no type check before binding defaultRenderer. (#2350)
The fix is simply to add a type check.

### DIFF
--- a/src/Engines/Custom.js
+++ b/src/Engines/Custom.js
@@ -170,7 +170,12 @@ class CustomEngine extends TemplateEngine {
       // Bind defaultRenderer to render function
       if ("then" in fn && typeof fn.then === "function") {
         // Promise, wait to bind
-        return fn.then((fn) => fn.bind({ defaultRenderer }));
+        return fn.then((fn) => {
+          if (typeof fn === "function") {
+            return fn.bind({ defaultRenderer });
+          }
+          return fn;
+        });
       } else if ("bind" in fn && typeof fn.bind === "function") {
         return fn.bind({ defaultRenderer });
       }
--- a/test/TemplateTest-CustomExtensions.js
+++ b/test/TemplateTest-CustomExtensions.js
@@ -466,3 +466,31 @@ test("Return undefined in compile to ignore #2267", async (t) => {
   let data = await tmpl.getData();
   t.is(await tmpl.render(data), undefined);
 });
+
+test("Return undefined in compile to ignore (async compile function) #2350", async (t) => {
+  let eleventyConfig = new TemplateConfig();
+  eleventyConfig.userConfig.extensionMap.add({
+    extension: "txt",
+    key: "txt",
+    compileOptions: {
+      cache: false,
+    },
+    getData: false,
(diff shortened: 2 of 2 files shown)


zachleat:
Thank you! This will ship with 2.0.0-canary.9. If you need it on 1.x, please let me know!

noelforte:
@zachleat it would be great to have this in a stable release, if 1.0.2 is planned to ship prior to v2, I'd love to make the upgrade on my project which is currently pinned to 1.0.0. As no one else has expressed interest in getting this fix sooner, I'm happy to wait; just wanted to voice my interest.

monochromer:
@zachleat i want this fix in 1.x too

PR #2367, opened 2022-05-06
fix liquid template argument parsing and shortcode undefined values

epelc:
- fixes undefined arguments in shortcodes and potentially other uses
- fixes #2348 and #2154

This fixes a major bug with liquid shortcodes not working in layout templates. They end up randomly getting `undefined` values for arguments after the first argument due to an order of execution issue with a changing `scope` context being sent to [evalValue](https://github.com/11ty/eleventy/compare/master...epelc:master#diff-cd8b0d5daef1e85d609f147e72a641e39dd0091c09dce6b8be1903e8dfd59a25R122).

I changed it from running the promise using await immediately to calling the promise immediately instead so it gets the correct current scope value. I then use `await Promise.all(pendingPromises)` to wait for each promise to resolve.

There is a separate example here which shows the differences in order of execution with calling the promise immediately with await(broken) vs calling it without await and then later awaiting all of the promises(fixed).

# Demo/why it fixes and didn't work before

> See example javascript below

 Notice how the regular version starts the two functions concurrently and their counter is then randomly iterated between the the functions. Consider each function an eleventy input file and the counter to be the scope in the liquid parser. You end up with the undefined arguments bug in this setup because the scope changes out of order compared to where the variables are being processed.

Compare with the fixed version under it in which the first file is completely executed then the next one. Nodejs is single threaded anyways so executing it in order synchronously while still using async/await is fine. We are just fixing the processing order here so that the counter/scope are processed in the corresponding order to their arguments.

```
> console.log('reg',await Promise.all([doConcurrentStuff(ctx),doConcurrentStuff(ctx)]))
Doing concurrent stuff
ctx { counter: 71 }
Doing concurrent stuff
ctx { counter: 72 }
ctx { counter: 73 }
ctx { counter: 74 }
ctx { counter: 75 }
ctx { counter: 76 }
ctx { counter: 77 }
ctx { counter: 78 }
ctx { counter: 79 }
ctx { counter: 80 }
reg [ [ 71, 73, 75, 77, 79 ], [ 72, 74, 76, 78, 80 ] ]
undefined
> console.log('fixed',await Promise.all([doConcurrentStufffixed(ctx),doConcurrentStufffixed(ctx)]))
Doing concurrent stuff with fixed passing of current value of ctx/scope
ctx { counter: 81 }
ctx { counter: 82 }
ctx { counter: 83 }
ctx { counter: 84 }
ctx { counter: 85 }
Doing concurrent stuff with fixed passing of current value of ctx/scope
[...]

PR #2369, opened 2022-05-06
add test for pr #2367

epelc:
I added a test for previous pr #2367 as per @zachleat.

It doesn't actually fail though on the old code as it only runs a single template file. I wasn't sure how to run multiple template files but this is everything else that's required. You need at least 2 pages to be rendered since this is an order of execution bug.

It does at least add a test for shortcodes which I think was missing anyways.

### DIFF
--- a/test/TemplateTest.js
+++ b/test/TemplateTest.js
@@ -320,6 +320,45 @@ test("One Layout (_layoutContent deprecated but supported)", async (t) => {
   t.is(data.keylayout, "valuelayout");
 });
 
+test("Liquid shortcode with multiple arguments(issue #2348)", async (t) => {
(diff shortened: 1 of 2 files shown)


zachleat:
Ah hey sorry, I’d recommend just adding a top level one like this https://github.com/11ty/eleventy/blob/9657b436afd29e3349946642cc624c2fa7f577dd/test/EleventyTest.js#L488-L502

then just add a few template files in a stubs folder

zachleat:
Fixed this! Got the test breaking and then it is fixed with the code from #2367.

Thank you again!

epelc:
Thanks @zachleat!

PR #2371, opened 2022-05-07
chore: Replace deprecated String.prototype.substr() with String.prototype.slice()

amareshsm:
String.prototype.substr() method is deprecated and no longer recommended. So replaced with String.prototype.slice() method.


zachleat:
Thank you! Shipping with `2.0.0-canary.10`

Issue #2424, opened 2022-06-06
Build fails for new markdown files using `git Last Modified` as the date


kareldonk:
**Describe the bug**

When a new markdown file has not yet been committed to the git repo and it is using `git Last Modified` as the date, and I try to build, I get the following error:

```
[11:watch    ] [11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11:watch    ] [11ty] mapA.date.getTime is not a function (via TypeError)
[11:watch    ] [11ty]
[11:watch    ] [11ty] Original error stack trace: TypeError: mapA.date.getTime is not a function
[11:watch    ] [11ty]     at sortFunctionDateInputPath (C:\Dev\Project\node_modules\@11ty\eleventy\src\Util\Sortable.js:107:17)
[11:watch    ] [11ty]     at Array.sort (<anonymous>)
[11:watch    ] [11ty]     at TemplateCollection.sort (C:\Dev\Project\node_modules\@11ty\eleventy\src\Util\Sortable.js:45:31)
[11:watch    ] [11ty]     at TemplateCollection.getAllSorted (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateCollection.js:25:17)       
[11:watch    ] [11ty]     at TemplateMap.getTaggedCollection (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateMap.js:513:32)
[11:watch    ] [11ty]     at TemplateMap.setCollectionByTagName (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateMap.js:262:44)
[11:watch    ] [11ty]     at TemplateMap.initDependencyMap (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateMap.js:284:20)
[11:watch    ] [11ty]     at async TemplateMap.cache (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateMap.js:333:5)
[11:watch    ] [11ty]     at async TemplateWriter._createTemplateMap (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateWriter.js:244:5)   
[11:watch    ] [11ty]     at async TemplateWriter.generateTemplates (C:\Dev\Project\node_modules\@11ty\eleventy\src\TemplateWriter.js:277:5)    
[11:watch    ] [11ty] Wrote 0 files in 3.98 seconds (v1.0.1)
```

**To Reproduce**

Add new file with date  `git Last Modified`  to existing project and build.

[...]

PR #2459, opened 2022-06-28
#2424 Fix failing builds for new markdown files using git Last Modified

Snapstromegon:
This PR fixes the issue where builds fail when you have at least two tempaltes where at least one is not yet in git and has the date set to "git last modified".

The issue was, that a timestamp as a number was returned in that case instead of a date object.


zachleat:
Oof, good catch! Thanks! 2.0.0-canary.13

PR #2463, opened 2022-06-29
Fix unittest execution on windows

Snapstromegon:
Unittests threw errors on windows because of some newline errors.
This happens with the most common git configs on windows.

This fix always checks for \n newlines in excerpts and that way fixes unittest execution on windows (which I use, so I mainly fixed it for myself)

### DIFF
--- a/src/TemplateContent.js
+++ b/src/TemplateContent.js
@@ -122,11 +122,18 @@ class TemplateContent {
       if (options.excerpt && fm.excerpt) {
         let excerptString = fm.excerpt + (options.excerpt_separator || "---");
         if (fm.content.startsWith(excerptString + os.EOL)) {
-          // with a newline after excerpt separator
+          // with an os-specific newline after excerpt separator
           fm.content =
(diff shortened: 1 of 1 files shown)


zachleat:
I’m curious: Windows is included in our test suite that runs automatically in GitHub actions: https://github.com/11ty/eleventy/actions

Why isn’t it failing there?

Snapstromegon:
I don't know how the setting on the github action version of windows is, but in my installation the following setting is set:
core.autocrlf=false

If this is defaulted to true on github actions, this might be the difference.

Snapstromegon:
Just checked it: https://github.com/Snapstromegon/gh-actions-playground/runs/7118885040?check_suite_focus=true

This seems to be the difference and there seems to be at least some people online who think that this should be false by default.

Anyways, this solution would make 11ty more compatible with the false option and shouldn't make a different anywhere else.

zachleat:
This will ship with 2.0.0-canary.19

PR #2464, opened 2022-06-29
#1028 #1029 Possible first implementation for additional default eleventy config files

### LINKED ISSUE #1028: By default search for .eleventy.cjs config file
If `.eleventy.cjs` exists, otherwise look for previous default `.eleventy.js`.

Related to #837.

### LINKED ISSUE #1029: By default look for eleventy.config.cjs and eleventy.config.js config files
Also think about a potentially breaking change for 1.0 as we want to move away from dot files and look for `eleventy.config.cjs` and `eleventy.config.js` too (file separately).

I think this will be an additional lookup instead of a replacement for `.eleventy.js`.

Snapstromegon:
Currently only .eleventy.js is supported as the default config file.

This PR implements a possible solution for multiple config file. It intentionally just implements support for multiple default config files.

The implementation might be a little flaky around unittests, but it passes all of them consistently.

A future PR should probably look into allowing usage of multiple config files at once. This only uses the first existing one.

fixes #1028
fixes #1029


zachleat:
We are going for it! Thank you @Snapstromegon!

I might reorder the file lookup array but it looked great!

zachleat:
v2.0.0-canary.15

zachleat:
Decided against `.eleventy.cjs`, reasoning here: https://github.com/11ty/eleventy/issues/1028#issuecomment-1216869727

Issue #1036, opened 2020-03-22
[Pagination] - introduce startPortion parametr or different chunk size for first paginated page


bodurkalukasz:
Hi, 
im trying to do a blog main page, where i will have 10 'latest' entries.

What i would like for pagination is to have this 10 entries on main page but then paginate for 24 entries. Is it possible with current setup? 

What comes to my mind at first is to have something like:

```
---
pagination:
  data: collections.posts
  startSize: 10
  size: 24
  alias: posts
---

```

Then, pagination would do that out of the box.

zachleat:
https://www.11ty.dev/docs/pagination/#the-before-callback does some of this but #4036 is the enhancement queue path forward here, please subscribe and upvote over there!

PR #2466, opened 2022-06-29
#1036 Add skip parameter to pagination plugin

Snapstromegon:
This PR adds an option to skip the first n items of a pagination.
This allows usecases as described in #1036, where one might want to show the first n items on onether page and then paginate over the rest.

### DIFF
--- a/src/Plugins/Pagination.js
+++ b/src/Plugins/Pagination.js
@@ -81,6 +81,12 @@ class Pagination {
 
     this.size = data.pagination.size;
     this.alias = data.pagination.alias;
+    this.skip = data.pagination.skip || 0;
+    if (typeof this.skip !== "number") {
+      throw new Error(
+        `Missing pagination skip in front matter data${this.inputPathForErrorMessages}`
+      );
+    }
     // TODO do we need the full data set for serverless?
     this.fullDataSet = this._get(this.data, this._getDataKey());
 
@@ -106,8 +112,8 @@ class Pagination {
         ];
       }
     } else {
-      // this returns an array
-      this.target = this._resolveItems();
+      // this returns an array and skips no elements by default
+      this.target = this._resolveItems().slice(this.skip);
 
       // Serverless Shortcut when key is not found in data set (probably running local build and expected a :path param in data)
       // Only collections are relevant for templates that don’t have a permalink.build, they don’t have a templateContent and aren’t written to disk
--- a/test/PaginationTest.js
+++ b/test/PaginationTest.js
@@ -883,3 +883,41 @@ test("Pagination and eleventyComputed permalink, issue #1555 and #1865", async (
   t.is(templates[1].data.page.url, "/venues/second/");
   t.is(templates[2].data.page.url, "/venues/third/");
 });
+
+test("Pagination with skip parameter, issue #1036", async (t) => {
+  let eleventyConfig = new TemplateConfig();
+  let tmpl = getNewTemplate(
+    "./test/stubs/pagination-skip.md",
+    "./test/stubs/",
+    "./dist",
+    null,
+    null,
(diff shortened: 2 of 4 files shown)


zachleat:
Huh! So this is like a start index?

Snapstromegon:
Yeah right. Basically it's a wrapper around `[].slice()`.

pdehaan:
Do we need test(s) for `skip` + the [`before` callback](https://www.11ty.dev/docs/pagination/#the-before-callback) as well?
Curious if the `skip` would occur before or after the `before` callback.

Snapstromegon:
@pdehaan I added only one happy path and one sad path test for skip in this PR.

Also the behavior is described in the docs PR: https://github.com/11ty/11ty-website/pull/1440
Basically it happens last, since it should take effect after reversing and filtering.

edwardhorsford:
Personally I'm unsure about the name 'skip'.

FWIW there's a few other related options pagination could have - start index, end index, count. I've had a few cases where I've wanted to limit or truncate my set of options. 

One possible way of supporting all of these could be for a `range` parameter.

Snapstromegon:
@edwardhorsford Ahh, that might be a good idea. I came from the point of the SQL "SKIP" command.
How would you make this range parameter?

My first suggestion would be something like this:
```yaml
range:
  first: 3
  last: -4
  count: 5
```
Where negative values mean "from last", where "-1" means the last element similar to [Array.at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).
Count is considered a max.
- If only first is set, it's like `[].slice(first)`.
- If first and count is set, it's like `[].slice(first, first+count)`
- If only first and last is set, it's like `[].slice(first, last+1)`
- If count is set, it's like `[].slice(0, count)`
- if last is set, it's like `[].slice(last)`
- if last and count is set, it's like `[].slice(last, last-count)`
- If all three are set, it's like `[].slice(first, last+1).slice(0, count)`

Additionally we could support a string value for range in the form of [Rust Ranges](https://doc.rust-lang.org/rust-by-example/flow_control/for.html#for-and-range), so `range: 0..5` or `range: 0..=4`. Here the first number is read as "first" from above and the second as "last" (for ..= or last-1 for ..).

edwardhorsford:
@Snapstromegon yep, I had a similar thought - using negative numbers to count backwards for the end.

I suspect your first example with an object and properties is easier to read and understand than the Rust ranges.

I might suggest different names though... perhaps:

```
range:
  start: 10
  end: -4
  limit: 100
```

I suspect users would be unlikely to want to use both `end` and `limit` simultaneously, but seems fine to support both as long as the priority is clear.

I speculate `count` might be misread as an expectation that number of pages should be generated - when in reality it's either a limit or truncation.

My main use for a limit so far has been where I'm working on a pagination file that outputs many hundreds of pages - but for development I only need a small number to test. My solution so far has been filtering the data with the `before` callback.

Snapstromegon:
@edwardhorsford `limit` sounds better. I also think of supporting the range and object syntax at the same time, so it's not one or the other (this is, so the simpler way of start and end can be shorter).

I also think that `limit` is fairly good semantically coming from SQL, since it clearly transports the priority (first select by start and end, then only take the limit amount).

What I personally disagree with, is that `start` and `end` are better names, because `first` and `last` clearly state wether or not the selected range is inclusive or not.

E.g. if I have this data:
```yaml
- zero
- one
- two
- three
- four
- five
- six
- seven
```

And I'd do
```yaml
range:
  start: 2
  end: 4
```

Would I get `['two', 'three']` or `['two', 'three', 'four']`?
(I would keep this similar to `[].slice()` and make the range non-inclusive.

PR #2467, opened 2022-06-29
Add passthrough logging filter #2198

### LINKED ISSUE #2198: Add passthrough logging filter
When debugging especially filter chains in templates it's often easier if you can just drop a logging filter in between to see the result on the console. 

Maybe it would be useful to include such a feature like the default log filter (or even an adaption of it) in the default config, which returns the input value.

**Option 1:**
Store the filter chain before the logging in a variable and then use it once for logging and once for the actual work

**Option 2:**
I normally add such a filter myself:
```js
eleventyConfig.addFilter("logging", (input, label, passthrough) => {
  console.log(`logging-${label}:`, input);
  if(passthrough) return input;
});
```

**Additional context**
This would help in cases where you have something like this in a template:
```njk
{{ someData | thatIsPassed | throughMultiple | filters | logging("Expected format?", true) | andStillUsed }}
```

I think switching the current log filter for the version above might not be compatible, because at the moment the direct binding to console.log allows passing any number of arguments, so it's either a breaking change, or needs to be a new filter.

Snapstromegon:
This change makes the `log` filter "passthrough" compatible, so you can put it in the middle of a filter chain.

That way you can use it like `{{ "Some Text" | log("Some logging") | upper }}`.

This fixes #2198.

### DIFF
--- a/src/defaultConfig.js
+++ b/src/defaultConfig.js
@@ -14,8 +14,8 @@ module.exports = function (config) {
     let pathPrefix = pathPrefixOverride || templateConfig.getPathPrefix();
     return urlFilter.call(this, url, pathPrefix);
   });
-  config.addFilter("log", (input, ...data) => {
-    console.log(input, ...data);
+  config.addFilter("log", (input, ...messages) => {
+    console.log(input, ...messages);
     return input;
   });
 
--- a/src/defaultConfig.js
+++ b/src/defaultConfig.js
@@ -14,7 +14,10 @@ module.exports = function (config) {
     let pathPrefix = pathPrefixOverride || templateConfig.getPathPrefix();
     return urlFilter.call(this, url, pathPrefix);
   });
-  config.addFilter("log", console.log);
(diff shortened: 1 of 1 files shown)


src/defaultConfig.js
```diff
@@ -14,7 +14,10 @@ module.exports = function (config) {
     let pathPrefix = pathPrefixOverride || templateConfig.getPathPrefix();
     return urlFilter.call(this, url, pathPrefix);
   });
-  config.addFilter("log", console.log);
+  config.addFilter("log", (input, ...data) => {
```
zachleat, line 17: Happy to merge this! Great idea.

Can we rename `data` to `messages` or `logMessages` or something with a bit more readability?

Snapstromegon: Great comment, let's call it messages, since that's more closely to how MDN calls it on console.log ([msg](https://developer.mozilla.org/de/docs/Web/API/Console/log)) and how typescript calls that parameter (messages).

I will update the PR.

PR #2477, opened 2022-06-30
Add tests to get UT Coverage for utils/Sortable to 100

Snapstromegon:
This PR contains some changes to the Sortable class, to make the API a little more consistent (IMO) while being backwards compatible.


zachleat:
2.0.0-canary.15 thanks!

Issue #1312, opened 2020-07-11
Issues with watching JS dependencies declared in configuration


zachleat:
Filed via @anandchowdhary

<img width="977" alt="image" src="https://user-images.githubusercontent.com/39355/87214792-7a214c00-c2f5-11ea-8737-b43956311f6c.png">

https://github.com/11ty/eleventy/issues/701#issuecomment-575613848

AurelienStebe:
Following up on this discussion from Discord : https://discordapp.com/channels/741017160297611315/741017160297611319/757230379202773002

The issue at hand on Discord is in a JS template, but I think it is the exact same issue mentioned here.
The "DeleteRequireCache" utility should probably delete the Module children cache entries also, using `require.cache[normalizedPath].children` in a recursive loop here : 
https://github.com/11ty/eleventy/blob/330191c10b9ab4b93a3abcf6406bbc350a47aa7d/src/Util/DeleteRequireCache.js#L8-L11

I could get a PR together, just checking for a green light and if such a fix is not already in the pipeline.

esharri2:
@AurelienStebe Did you develop a solution or workaround for this? I'm having a few issues with changes to required JS files not triggering a refresh (or in some cases requiring that I restart the server). I'm seeing it with files required in the config, and files required in layouts.

AurelienStebe:
No, this was just a drive-by bug fix. I had the exact same issue in my own project the week before, so I knew the solution.

It should really just need a few lines of code : before deleting the module itself from the require cache, just loop on its children and call the function recursively. Unfortunately, it needs to be fixed in this very file in Eleventy, I don't think you can do much from outside (apart from deleting the whole cache maybe, but that's about the same as restarting the process). This should make sure that dependencies of dependencies are re-read from disk in "live" mode, however I don't know if they are all "watched" correctly (but there is `eleventyConfig.addWatchTarget()` for that).

I thought work was being done on the "--watch" mode and incremental builds, so I didn't move forward at the time.

tomglynch:
@zachleat is this in the pipeline or worth me getting a PR together?

tomglynch:
@AurelienStebe after editing the code in eleventy/src/Util/DeleteRequireCache.js to look for the children modules and delete the cache for them also, it still does not re

> Yeah I just hit the same problem. Any dependencies of an `.11ty.js` layout are not correctly updated when the page reloads. This makes it almost impossible to work with JS templates :/
> 
[...]

PR #2479, opened 2022-07-01
#1312 Improve situation for dependency monitoring

Snapstromegon:
This change adds a "PathNormalizer" with tests which converts all paths to be unix compatible (replace "\" with "/"). That way auto reloading dependencies of the config also works on windows.

Sadly nested dependencies still always lag exactly one update behind, so if you have the following dependency graph:
- .eleventy.js depends on
- middle.js depends on
- nested.js

and you update nested, you see the following:
- do update 1 -> no result
- do update 2 -> see update 1
- do update 3 -> see update 2

During a short lookaround, I was unable to resolve this issue.

This should improve the situation around #1312

### DIFF
--- a/src/EleventyWatch.js
+++ b/src/EleventyWatch.js
@@ -1,4 +1,5 @@
 const { TemplatePath } = require("@11ty/eleventy-utils");
+const PathNormalizer = require("./Util/PathNormalizer.js");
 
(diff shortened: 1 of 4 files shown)


zachleat:
Thank you!

Issue #2287, opened 2022-03-25
👍👎 Tag like syntax for `eleventyExcludeFromCollections` to exclude page from specific tag's collection


nitin2953:
https://discord.com/channels/741017160297611315/872911930120110190/956590310987792414

<details>
<summary>As discussed in above discord thread</summary>

> My file structure
> ```
> blog
>  ├─ 1st-post.md
>  ├─ 2nd-post.md
>  ├─ 3rd-post.md
>  ├─ blog.json
>  └─ index.njk
> ```
> `blog.json`
> ```json
> {
>     "layout": "post.njk",
>     "tags": "blog_post"
> }
> ```
> `index.njk`
> ```hbs
> {%- for post in collections.blog_post -%}
>     <h2>{{ post.data.title }}</h2>
>     <p>{{ post.data.description }}</p>
>     <a href="{{ post.url }}">Read more</a>
> {% endfor -%}
> ```
> `_site/blog/index.html`
> ```html
> <h2>Blog</h2> ◀◀◀
> ...
> <h2>1st Post</h2>
> ...
> <h2>2nd Post</h2>
> ...
> <h2>3rd Post</h2>
> ...
> ```
> 
> **Problem:**
> I want to exclude `blog/index.njk` ONLY form `collections.blog_post`
> I can't set `eleventyExcludeFromCollections: true` since I want it in `sitemap.xml` which is using `collections.all`
> 
> **Solution:**
> New option for `eleventyExcludeFromCollections` with syntax like https://11ty.dev/docs/collections/#tag-syntax.
> e.g. `eleventyExcludeFromCollections: blog_post` for excluding `blog/index.njk` ONLY from `collections.blog_post`
> 
> **Edit 1: Another Solution**
> Moving `blog/index.njk` to parent folder (and adding `permalink: /blog/` in front matter) will stop adding `blog/blog.json`'s tag to it & it won't include it in it's collection.
> 
> **Edit 2:**
> BUT this feels hacky & will mess files structure in future.

</details>

 
**Problem**
- `eleventyExcludeFromCollections: true` remove page from all collections including `collections.all`.
- I want to exclude `index.njk` file (which is residing with `<dir>.json` file) ONLY from `collections.blog_post` OR any specific collection except `collections.all` which is used `sitemap.xml`.

**Solution**
**New option for `eleventyExcludeFromCollections` with [tag like syntax](https://11ty.dev/docs/collections/#tag-syntax).**
[...]

Issue #1175, opened 2020-05-12
Allow eleventyExcludeFromCollections to take an array of collections


paulrobertlloyd:
**Is your feature request related to a problem? Please describe.**
I have an index file alongside a group of posts

```
posts
├─ 01-01-2020-first-post.md
├─ 02-02-2020-second-post.md
├─ index.njk
└─ posts.11tydata.js
```

`posts.11tydata.js` adds a `post` tag to all posts in this folder, and `index.njk` loops through `collection.post`, but this loop includes `index.njk` as well as all the individual posts.

I can exclude the index by using `eleventyExcludeFromCollections: true`, but that excludes it from all collections. But in my `sitemap.xml` I want to include it when I use `collection.all`.

**Describe the solution you'd like**
Allow `eleventyExcludeFromCollections` to take an array of collections to be excluded from:

```yaml
eleventyExcludeFromCollections:
- post
- foo
- bar
```

**Describe alternatives you've considered**

Currently I add `tags: false` to the index file, but it’s not as explicit.

zachleat:
I like this idea! It must go into the queue!

> This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

> View [the enhancement backlog here](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+). Don’t forget to upvote the top comment with 👍!

davidhund:
Apologies if commenting here (and now) is inappropriate. 
I'm not entirely sure how the "closed-but-queued issues" thing works. ;-)

Anyway: this gets a strong 👍  from me: Paul's use-case is something I also run into all the time…

Ryuno-Ki:
@davidhund Basically, Zach's looking at https://github.com/11ty/eleventy#major-roadmapped-features for making decisions.

architchandra:
Please add the ability to exclude from collections on a per-collection basis (or include in `collections.all` while excluding from others). It will be really helpful.

nitin2953:
Please reopen this issue 😥,
without this feature my directory structure is messing up 😭.

zachleat:
#2482 was merged—this will ship with 3.0.

zachleat:
Docs shipping to https://www.11ty.dev/docs/collections/#how-to-exclude-content-from-collections

PR #2482, opened 2022-07-01
#1175 and #2287 Allow eleventyExcludeFromCollections to take a list of tags

Snapstromegon:
This is a first implementation for #1175 and #2287 to gather feedback.

This change allows for more fine grained control over how a template is excluded from collections by allowing to set eleventyExcludeFromCollections to a tag or list of tags to exclude the template from.

This is meant as a first implementation for feedback. Also I'm not completely sure if excluding the `all` Collection is done correctly.

I only added minimal tests, but probably some more are needed.


zachleat:
Conflicts fixed and this will ship with 3.0.

Issue #2436, opened 2022-06-14
Change default `node_modules` ignores


zachleat:
The `eleventy-base-blog` build fails after you add a Netlify plugin. Here’s some build log output from the eleventy-base-blog project but would likely be the similar for any Eleventy 0.x, 1.x, or 2.x project:

```
0:20:59 AM: [11ty] Problem writing Eleventy templates: (more in DEBUG output)
10:20:59 AM: [11ty] 1. Having trouble rendering njk template ./.netlify/plugins/node_modules/accepts/README.md (via TemplateContentRenderError)
10:20:59 AM: [11ty] 2. "sh" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight (via Error)
10:20:59 AM: [11ty]
10:20:59 AM: [11ty] Original error stack trace: Error: "sh" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight
10:20:59 AM: [11ty]     at module.exports (/opt/build/repo/node_modules/@11ty/eleventy-plugin-syntaxhighlight/src/PrismLoader.js:20:11)
10:20:59 AM: [11ty]     at Object.highlight (/opt/build/repo/node_modules/@11ty/eleventy-plugin-syntaxhighlight/src/markdownSyntaxHighlightOptions.js:25:35)
10:20:59 AM: [11ty]     at Object.default_rules.fence (/opt/build/repo/node_modules/markdown-it/lib/renderer.js:53:27)
10:20:59 AM: [11ty]     at Renderer.render (/opt/build/repo/node_modules/markdown-it/lib/renderer.js:332:38)
10:20:59 AM: [11ty]     at MarkdownIt.render (/opt/build/repo/node_modules/markdown-it/lib/index.js:544:24)
10:20:59 AM: [11ty]     at /opt/build/repo/node_modules/@11ty/eleventy/src/Engines/Markdown.js:74:38
10:20:59 AM: [11ty]     at processTicksAndRejections (node:internal/process/task_queues:96:5)
10:20:59 AM: [11ty]     at async Template._render (/opt/build/repo/node_modules/@11ty/eleventy/src/TemplateContent.js:432:22)
10:20:59 AM: [11ty]     at async Template.getTemplateMapContent (/opt/build/repo/node_modules/@11ty/eleventy/src/Template.js:1077:19)
[...]

PR #2492, opened 2022-07-12
Ignore nested node modules by default

nasivuela:
It changes Eleventy default ignore from `node_modules/**` to `**/node_modules/**`.
https://github.com/11ty/eleventy/issues/2436

Docs section that will need to be updated => https://www.11ty.dev/docs/ignores/#node_modules


zachleat:
It’s happening! Thank you @nasivuela!

Will ship with Eleventy 2.0.0-canary.15

Issue #1297, opened 2020-07-03
Tags doesn't work correctly inside eeleventyComputed


winston0410:
**Describe the bug**

Tags doesn't work correctly inside `eleventyComputed`.  The collections of that tag returns undefined.

**To Reproduce**

1. Add tags inside front matter
```
---
eleventyComputed:
  tags: ['cv', 'menu-{{index.lang}}']
  permalink: "/{{ index.lang | urlEncode}}/cv/"
---
```
2. Get the value of the collection with `log`
{{ collections.cv | log }}
3. The collection return undefined.

**Expected behavior**

I expect tags being used inside `eleventyComputed` can be functional.

**Screenshots**

**Environment:**

- OS and Version: [e.g. Windows/Mac/Linux] 
- Eleventy Version [via `eleventy --version` or `npx @11ty/eleventy --version`] 0.11.0

**Additional context**

zachleat:
Correct, we have a pretty hefty warning at the top of https://www.11ty.dev/docs/data-computed/ about this

PR #2540, opened 2022-08-29
#1297 Make tags work correctly inside eleventyComputed

stopnoanime:
While making a page myself I came across the same issue as described here: #1297 so I tried to fix it.

My solution works for the test cases I tried and doesn't break anything else according to the tests, but I would like someone else more experienced in the codebase to take a look at it and see if it's a good way to fix this issue.

I would like to see this fixed, but I understand if I turns out to be unfeasible or something not in the current scope.

---

Some more info about my solution:
- If there is no `tags` property in `eleventyComputed`, continue with the normal flow
- If there is a `tags` property in `eleventyComputed`, skip adding tags to graph in `getMappedDependencies` and instead do it in `getDelayedMappedDependencies` when the data from `eleventyComputed` has already been merged to entry data.
- I also added a test case

### DIFF
--- a/src/TemplateMap.js
+++ b/src/TemplateMap.js
@@ -128,7 +128,7 @@ class TemplateMap {
         // collections.all
         graph.addDependency(tagPrefix + "all", entry.inputPath);
 
-        if (entry.data.tags) {
+        if (entry.data.tags && !entry.data.eleventyComputed?.tags) {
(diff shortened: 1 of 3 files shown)


stopnoanime:
I just realized this soulution doesn't work for all cases so I am closing this PR

Issue #2560, opened 2022-09-10
i18n plugin not working with pagination that creates multiple pages


rileymacisaac:
**Describe the bug**
I have one file for each language (en/fr) that creates all the pages on my site from data from Storyblok. These files are `en/story.njk` and `fr/story.njk`.

I'm trying out the new i18n plugin and realized that the alternates functionality does not work with pagination as I'd imagine.

It seems that the plugin tries to make alternates for each template, rather than each url. 

For example, `en/story.njk` creates 10+ pages, but the plugin only creates alternates for the **FIRST** page built by eleventy and then stops.

**Expected behavior**
I'd expect the alternates to be created for each matching url (with the lang code removed).

**Environment:**
Eleventy Version 2.0.0-canary.15

I'll try to add more details and continue to investigate a way forward.

rileymacisaac:
Related: https://github.com/11ty/eleventy/commit/5e36ca3c71f4c952ac567961493122e6feb8da84

stevygee:
I think I have the same issue: When creating multiple pages from data using pagination, the `locale_links` filter only returns the alternates on the first page and doesn't work for the pages that follow.

rileymacisaac:
> I think I have the same issue: When creating multiple pages from data using pagination, the `locale_links` filter only returns the alternates on the first page and doesn't work for the pages that follow.

@stevygee you should check out the merge request I made (#2567) and let me know if that works for your use case. I'm not using it yet of course but I'd be interested to see what you think.

stevygee:
@rileymacisaac Finally had a chance to try your fork, unfortunately it doesn't solve my issue.

jmzagorski:
I have the same problem, so I am glad I found this issue because I thought I was doing something incorrect. I have 11ty.js and markdown templates that use global data to generate pages. Global data is fetched from a server and that data is read to know how many pages to build. The `locale_links` only works for a few pages, but the majority of pages are empty. 

I tried @rileymacisaac solution and it works for my templates that do not get generated by server-side database data. These would be my "docs" page that have markdown text in it. I inspected `contentMaps` in the event `eleventy.contentMap` and saw `localeUrlMaps` had 2 objects for each page (which I am guessing are the two locale pages, but I did not dive into the code), but for some reason my auto generated pages still return an emtpy array for `locale_links`.

If my users want this feature, maybe I will investigate the code a bit deeper, just wanted to +1 this issue with another report. Great work on the plugin!

```
_includes/
├── en
[...]

PR #2567, opened 2022-09-13
Change i18n alternates to work off of url, rather than pagination tem…

rileymacisaac:
This is probably not ready for release, and I'm not sure if it matches the original intentions of the i18n plugin, but this works for me on my tests and my needs.

#2560

### DIFF
--- a/src/Plugins/I18nPlugin.js
+++ b/src/Plugins/I18nPlugin.js
@@ -17,6 +17,17 @@ class LangUtils {
       .find((entry) => Comparator.isLangCode(entry));
   }
 
+  static removeLanguageCodeFromUrl(url) {
+    let langCode = (url || "")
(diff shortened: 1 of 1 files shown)


zachleat:
Hmm, I’m not sure this is aligned with the requirements of the i18n plugin. Alternates must use the file name rather than the url because the url can be localized. For example, urls at `/es/cuarta-publicación/` and `/en/fourth-post/` can match to the same parent

zachleat:
> This is an automated message to let you know that this issue requires additional input from the original poster and for the health of the repository issue tracker the issue will be closed. This is to help alleviate issues hanging open waiting for a response from the original poster.

> Do not let the issue’s closing deter you. After you’ve supplied the required information, the issue will be reopened. Thanks!

PR #2575, opened 2022-09-23
esm

bennypowers:
This PR converts the Eleventy source to ESM.

The main impediment to merging are the features relying on the dependency graph.

In order to allow async work when instantiating objects (typically dynamic `import()` replacing synchronous `require()`), most constructor calls have been replaced with static async methods. 
```js
const foo =
  // new Foo(a, b, c);
  await Foo.from(a, b, c);
```

This branch does not provide a cjs build, but one could be generated.


src/Eleventy.js
```diff
@@ -33,7 +39,13 @@ const eventBus = require("./EventBus");
  * @param {String} output - Where to write rendered files to.
  * @returns {module:11ty/eleventy/Eleventy~Eleventy}
  */
-class Eleventy {
+export default class Eleventy {
+  static async from(input, output, options = {}, eleventyConfig = null) {
```
j-f1, line 43: fwiw you can make a constructor async with this pattern:

```js
export default class SomeClass {
  constructor() {
    return (async () => {
      // do something async...
      return this;
    })();
  }
}
const someObject = await new SomeClass();
```

It is unfortunately not compatible with TypeScript, though: https://github.com/Microsoft/TypeScript/issues/29138

bennypowers: very cool. nice pattern, and in this PR i think .from is better because it opens the door for future TS support. thanks for the tip, very cool

zachleat:
Appreciate this PR! ESM is slated to ship with 3.0 and we’re working on finishing 2.0 in the next month or two!

bennypowers:
Is there interest in rebasing this, or do maintainers prefer to start from scratch?

bennypowers:
Closing because jewish

PR #2577, opened 2022-09-26
Import EleventyShortcodeError into RenderPlugin.js

j-f1:
This fixes the following error, which obscures the actual issue causing an error to be thrown.

```
[11ty] Unhandled rejection in promise: (more in DEBUG output)
[11ty] EleventyShortcodeError is not defined (via ReferenceError)
[11ty] 
[11ty] Original error stack trace: ReferenceError: EleventyShortcodeError is not defined
[11ty]     at /path/to/project/node_modules/@11ty/eleventy/src/Plugins/RenderPlugin.js:225:21
[11ty]     at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### DIFF
--- a/src/EleventyShortcodeError.js
+++ b/src/EleventyShortcodeError.js
@@ -1,0 +1,5 @@
+const EleventyBaseError = require("./EleventyBaseError");
+
+class EleventyShortcodeError extends EleventyBaseError {}
+
+module.exports = EleventyShortcodeError;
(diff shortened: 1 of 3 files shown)


zachleat:
Thank you! This will ship with 2.0.0-canary.19. I did fix this also in an earlier canary, but your code was a little cleaner 🏆

Issue #2591, opened 2022-10-06
`pathPrefix` is overridden in ^2.0.0-canary builds


kev4ev:
**Describe the bug**
The `canary` builds of eleventy are ignoring the `pathPrefix` set in the .eleventy.js config file. I have remediated the issue in [PR 2592](https://github.com/11ty/eleventy/pull/2592) for your review. Thank you in advance for your time and the awesomeness that is `eleventy`.

**To Reproduce**
I have an `.eleventy.js` config file that exports the following: 

```js
module.exports = function(eleventyConfig){
    return {
        dir: {
            input: 'src',
            output: 'docs'
        },
        markdownTemplateEngine: 'njk', 
        htmlTemplateEngine: 'njk', 
        pathPrefix: '/myprefix/'
    }
}
```
When I run eleventy...
`$ eleventy --serve` 

* the path prefix is still set to `/` and the dev server's root is `/`
* links piped through `url` do not receive the path prefix

I have even tried adding the new [HTML base plugin](https://www.11ty.dev/docs/plugins/html-base/) to my config, but still no path prefixes.

**Expected behavior**
The pathprefix provided in the config file should be used as the base url for the dev server and transformed on links piped through `url` filter or transformed by `HTML base` plugin. 

**Environment:**

- OS and Version: Linux and Windows
- Eleventy Version `2.0.0-canary.16`

**Additional context**
Updating the `TemplateConfig.setPathPrefix` method as such fixes the issue and does not cause any obvious negative side effects from the functional testing I have performed: 
```js
setPathPrefix(pathPrefix) {
    if(pathPrefix && pathPrefix !== "/"){
      debug("Setting pathPrefix to %o", pathPrefix);
      this.overrides.pathPrefix = pathPrefix;
    }
  }
```

kev4ev:
Hi @zachleat just wondering if anyone on the team has had a chance to take a look at this. I'm sure your very busy but any help you can provide would be greatly appreciated as I'm currently working on a `2.0`-specific plugin. Thanks!

zachleat:
Mergin’ thank you! Will ship with 2.0.0-canary.19.

For visitors currently on previous versions of the 2.0 canaries, note that `--pathprefix` on the command line does still work!

PR #2592, opened 2022-10-06
#2591 Fix for path prefix in canary build

kev4ev:
Fixes issue #2591 where `canary` builds are disregarding path prefixes set in .eleventy.js config file.


zachleat:
Prompted and prodded by @stuartlangridge for this one

zachleat:
This will ship with 2.0.0-canary.19.

Thank you!

Issue #2226, opened 2022-02-18
large data sets: 1.x has issues that 0.11.1, 0.12.1 do not


SignpostMarv:
**Describe the bug**
I've a semi-open site generator project that squishes gigabytes of data sources down to about 6.9k pages to be processed by eleventy in two ways:

1) the legacy markdown repo
2) the more up-to-date json data source (pagination ftw!)

0.11.1 handles the 6.9k documents & 15.6mb json file without issue, 1.0.0 falls over in a similar fashion to that described in #695

**To Reproduce**
The site generator is semi-open in that the source is available at https://github.com/Satisfactory-Clips-Archive/Media-Search-Archive, but it's not feasible to stash 2.7gb+ of source data into the repo, so the repro steps aren't readily reproduceable by anyone that doesn't have the data set.

While the method mentioned in #695 of specifying `--max-old-space-size` does move the goalposts somewhat, it still falls over with 8gb assigned.

Steps to reproduce the behaviour:

1. `npm run build` or `./node_modules/.bin/elevent --config=./.eleventy.pages.js`
2. watch & wait

**Expected behaviour**
1.x to handle 6.9k markdown documents or 6.9k json data file entries as reliably as 0.11.x does

**Screenshots**

```
<--- Last few GCs --->

[16544:000001D55D359230]   168029 ms: Mark-sweep 4038.5 (4130.3) -> 4024.3 (4130.3) MB, 3347.8 / 0.0 ms  (average mu = 0.138, current mu = 0.017) task scavenge might not succeed
[16544:000001D55D359230]   171431 ms: Mark-sweep 4039.7 (4131.5) -> 4025.8 (4132.0) MB, 3345.7 / 0.0 ms  (average mu = 0.081, current mu = 0.016) task scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF70641DF0F v8::internal::CodeObjectRegistry::~CodeObjectRegistry+113567
 2: 00007FF7063AD736 v8::internal::MicrotaskQueue::GetMicrotasksScopeDepth+67398
 3: 00007FF7063AE5ED node::OnFatalError+301
 4: 00007FF706DA0CAE v8::Isolate::ReportExternalAllocationLimitReached+94
 5: 00007FF706D8B2FD v8::Isolate::Exit+653
[...]

PR #2633, opened 2022-10-30
RFC: #2627: Fix template write EMFILE when writing a large number of pages

### LINKED ISSUE #2627: EMFILE: too many open files when attempting to render a large (170000) number of pages (Eleventy 2.0.0-canary.16)
**Describe the bug**
EMFILE: too many open files when attempting to render a site with a large number of pages from a large data set.

When it is triggered, it is triggered via EleventyTemplateError.

The number of files being opened is finite: on a system with raised limits it successfully builds.

I have one 40MB json file, containing an array of about 170000 objects, and one template that generates one page per entry. ([The site](https://github.com/kisaragi-hiu/kemdict/tree/e2627) is supposed to be a static dictionary.)

Eleventy version: 2.0.0-canary.16. The site only builds on 2.0 due to #2360.

This behavior has been documented before in <https://github.com/11ty/eleventy/issues/2226#issuecomment-1120188230>.

Personally, I think this is an issue that can be fixed, as there is no reason why the output files, which depend only on the input file and not on each other, should all be held open at the same time.

**To Reproduce**

Edit: An earlier version said it took an hour to see the error. That is incorrect: I misread my CI logs, and Eleventy only took 3 minutes.
Edit: This has been rewritten to be more exact and to use error messages from the site generated during the reproduction.

1. Create a new test folder
2. Generate the test data and put it in `_data/a.json`, containing an array with a large number of objects, with each object having a unique, say, "title" key. I use Emacs for this:

   ```sh
   mkdir -p _data
   cat > x.el <<"HERE"
   (require 'json)
   (let (lst)
     (dotimes (i 200000)
       (push (let ((tmp (make-hash-table :test #'equal)))
               (puthash \"title\" (format \"%s\" i) tmp)
               tmp)
             lst))
     (with-temp-file \"_data/a.json\"
       (insert (json-encode lst))))
   HERE
   emacs --script x.el
   ```

   The JSON looks something like this:

   ```json
   [{"title": "199999"}, {"title": "199998"}, ...]
   ```

3. Create one template to generate pages for each item:

   ```markdown
   ---
   pagination:
     data: a
     size: 1
     alias: item
   permalink: "item/{{ item.title }}/index.html"
   ---
   {{ item.title }}
   ```

3. `npx @11ty/eleventy --quiet` (`--quiet` is to silence the list of rendered files)
4. On a system with higher FD limits, this will succeed.
5. On a more constrained system, an EMFILE error will be raised (from EleventyTemplateError).

   On Linux the FD limit can be lowered for a command like this:

   ```shell
   prlimit --nofile=50000 npx @11ty/eleventy --quiet
   ```

```
$ prlimit --nofile=50000 npx @11ty/eleventy --quiet
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble writing to "_site/item/199999/index.html" from "./a.njk" (via EleventyTemplateError)
[11ty] 2. EMFILE: too many open files, open '_site/item/49141/index.html' (via Error)
[11ty] 
[11ty] Original error stack trace: Error: EMFILE: too many open files, open '_site/item/49141/index.html'
[11ty] Wrote 0 files in 93.93 seconds (v2.0.0-canary.16)
```

**Expected behavior**
Writing many files from one template should not cause an EMFILE error.

**Environment:**

Where the error is encountered:

- OS and Version: Ubuntu 20.04 (on GitHub Actions), with an FD limit of somewhere between 1000 and 20000
- Eleventy Version: 2.0.0-canary.16

Where the error is not encountered:

- OS and Version: Arch Linux, with an FD limit in the millions
- Eleventy Version: 2.0.0-canary.16

**Additional context**

I can see that in TemplateWriter.js (generateTemplates) there is essentially an async, unbuffered loop to _generateTemplate → generateMapEntry (Template.js) → _write → writeFile (from graceful-fs).

(This is the source of the EMFILE because it's the only place where EleventyTemplateError can be raised.)

This would rapidly exhaust file descriptors if graceful-fs weren't used. But graceful-fs *is* used, and it should have handled this, and I don't understand why it's happening.

Perhaps Eleventy should also queue up file writes?

kisaragi-hiu:
This fixes #2627. Caveats below, though.

When paginating a data file with a large number of pages, for some reason despite using graceful-fs when given a large number of files it would still end up throwing EMFILE. The error is emitted through TemplateWriter.js (generateTemplates), in an async, unbuffered loop to _generateTemplate → generateMapEntry (Template.js) → _write → writeFile (from graceful-fs).

```
$ prlimit --nofile=50000 npx @11ty/eleventy --quiet
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble writing to "_site/item/199999/index.html" from "./a.njk" (via EleventyTemplateError)
[11ty] 2. EMFILE: too many open files, open '_site/item/49141/index.html' (via Error)
[11ty] 
[11ty] Original error stack trace: Error: EMFILE: too many open files, open '_site/item/49141/index.html'
[11ty] Wrote 0 files in 93.93 seconds (v2.0.0-canary.16)
```

This PR limits the number of concurrent writes on Eleventy's side to completely avoid hitting the FD limit.

##### Caveats

I've set the maximum number of concurrent writes to 1000, just below 1024 which some systems use as the default FD limit. Actually reading the current process FD limit would be much better, as this causes an unnecessary slow down for projects that go above 1000 concurrent writes but stays below the actual FD limit. I don't know if there is a standard way of reading the process FD limit that doesn't involve shelling out.

Also, as shown below, for some reason using 1000 as the maximum number of concurrent writes still allows it to build without hitting EMFILE even if the FD limit is lowered to 512. My hypothesis is that this is a low enough number that it doesn't overwhelm graceful-fs.

##### Results

Without this patch, a 80000-entry project generated with https://github.com/SignpostMarv/11ty-eleventy-issue-2226/commit/2e8f2da fails to build with EMFILE, as documented by https://github.com/11ty/eleventy/issues/2226#issuecomment-1120188230.

```
$ git clone https://github.com/SignpostMarv/11ty-eleventy-issue-2226 -b eleventy-2.0
$ cd 11ty-eleventy-issue-2226
$ npm install
$ node generate.js 80000
$ # using prlimit to explicitly set the FD limit to 1024 so this command can be run without checking system configuration
$ prlimit --nofile=512 npx @11ty/eleventy
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble writing to "./output/2.0.0-canary.16/80000/92e08c830313b35e9cc22f85492058672e3429bd250cdc99e9c4292408feb95d/7cc39edb-e5ae-4bfc-bef5-d613277ba061.html" from "./11ty/pages/test.njk" (via EleventyTemplateError)
[11ty] 2. EMFILE: too many open files, open './output/2.0.0-canary.16/80000/92e08c830313b35e9cc22f85492058672e3429bd250cdc99e9c4292408feb95d/2acfb91a-d3a0-4429-a6a5-e67157b9fa8b.html' (via Error)
[11ty] 
[11ty] Original error stack trace: Error: EMFILE: too many open files, open './output/2.0.0-canary.16/80000/92e08c830313b35e9cc22f85492058672e3429bd250cdc99e9c4292408feb95d/2acfb91a-d3a0-4429-a6a5-e67157b9fa8b.html'
[11ty] Wrote 0 files in 310.85 seconds (v2.0.0-canary.16)
```

With this patch:

```
$ prlimit --nofile=512 npx @11ty/eleventy --quiet
[11ty] Wrote 80000 files in 28.70 seconds (0.4ms each, v2.0.0-canary.16)
```

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -58,6 +58,7 @@ class Template extends TemplateContent {
 
     this.extensionMap = extensionMap;
 
+    this._availableConcurrentWrites = 1000;
     this.linters = [];
     this.transforms = [];
     this.templateData = templateData;
@@ -767,9 +768,25 @@ class Template extends TemplateContent {
         );
       }
 
+      // Despite using graceful-fs, for some reason when given a large
+      // number of files it would still end up throwing EMFILE. Limit
+      // the number of concurrent writes to avoid that.
+      await new Promise((resolve) => {
+        let check = () => {
+          if (this._availableConcurrentWrites > 0) {
+            resolve();
+          } else {
+            setTimeout(check, 1000);
+          }
+        };
+        check();
+      });
+      this._availableConcurrentWrites--;
+
       return writeFile(outputPath, finalContent).then(() => {
         templateBenchmark.after();
         this.writeCount++;
+        this._availableConcurrentWrites++;
         debug(`${outputPath} ${lang.finished}.`);
         return {
           inputPath: this.inputPath,


zachleat:
HMM, this looks great!

I’m curious: the concurrency check implemented here is template-specific, which means that it will only fix EMFILE issues stemming from a one-to-many pagination template (not from many disparate source templates). Do we want a global concurrency check instead?

zachleat:
> This module cannot intercept or handle EMFILE or ENFILE errors from sync methods. If you use sync methods which open file descriptors then you are responsible for dealing with any errors.

For myself, I would also note the above quoted from the graceful-fs docs: https://www.npmjs.com/package/graceful-fs

Just as a sanity check, we are using a promisified writeFile function from graceful-fs: https://github.com/11ty/eleventy/blob/4e8f1a8a0c067f15b6f8e397d3bd70de5ba2f6be/src/Template.js#L770 so I think we’re okay here.

I’m curious if we should be filing this upstream in `graceful-fs` instead? I see some similar issues but most of the relevant ones are closed as fixed.

https://github.com/isaacs/node-graceful-fs/issues?q=is%3Aissue+EMFILE+sort%3Aupdated-desc

zachleat:
Just thinking out loud here I wonder if the comment I left here is valid or not: https://github.com/isaacs/node-graceful-fs/issues/160#issuecomment-997284720 Maybe using `util.promisify` with `graceful-fs` isn’t working as expected!

zachleat:
I think the other possibility here is that this error might be specific to addPassthroughCopy and that graceful-fs is working fine for template writing?

AramZS:
Digging into this approach and testing it out, but wanted to say that practically the file limit should likely be set to whatever `ulimit` is on the host system? Or whatever the version of that is for Windows, no?

AramZS:
I confirm that this works in both the build and serve contexts it can be run in. This is strange to me and perhaps implies that there is likely a deeper problem in how Eleventy is handling things? I'm not sure.

AramZS:
Thinking that this might be an issue with the use of `fs` instead of `graceful-fs` elsewhere I've tried to flip more `fs` calls to `graceful-fs`. (Test coverage not completed, but you can see what I'm doing here - https://github.com/11ty/eleventy/compare/master...AramZS:eleventy:switch-more-to-graceful-fs )

I think this would cover most of the possible `fs` calls, but it did not resolve the issue like the PR here did.

AramZS:
I did some further experimentation with running this over a longer period of `--serve` and found that, interestingly, this worked initially but given enough changes done in a row Eleventy will still crash out with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`. I'll fiddle with the memory setup to see if it will also have the EMFILE crash occur given enough memory during the `--serve` rebuild.

AramZS:
Continuing to test and @zachleat I am starting to think the same as you are here, that perhaps `util.promisify` is doing something unexpected in this process. 

Also,w orth noting that [it appears the OSX max files open is 256](https://superuser.com/questions/827984/open-files-limit-does-not-work-as-before-in-osx-yosemite).  When I checked the setting on my OSX machine, 256 was still the default (12.6.1)

kaleb:
I solved this in my own build by patching [graceful-fs timeout](https://github.com/isaacs/node-graceful-fs/blob/1cc1135e76f84b45981e18438ce258ab18a051a1/graceful-fs.js#L418) to be greater than `60000`. It is not ideal, but it works for me currently.

epelc:
@AramZS @zachleat if you take a look at my recent PR I've switched all sync and fs.promise api calls to use graceful-fs async api. Should be a good step forward before potentially adding this one.

Graceful-fs can't catch the fs.promise calls and also doesn't handle sync.

https://github.com/11ty/eleventy/pull/2633

zachleat:
@epelc I have merged your PR! I think we’ll roll with that approach, as graceful-fs manages simultaneous writes internally (see https://github.com/isaacs/node-graceful-fs/issues/3)

zachleat:
Putting this one on-hold for now as we investigate the efficacy of #2915 to solve this one (currently on `main`, shipping with 3.0.)

timmywil:
FWIW, I still run into this with large builds on Windows using eleventy@ 4501f5884f7914e6618ae0589685147f31e0afb7.

It may still be worth considering this solution as well.

epelc:
@timmywil did you add the graceful da patch at top of your eleventy file. That’s required in addition to my patch. 

https://github.com/11ty/eleventy/pull/2915

Issue #2780, opened 2023-01-28
`compileOptions.permalink` default changed from `true` to `"raw"` for Custom Template Syntaxes


danburzo:
### Operating system

macOS Ventura 13.1

### Eleventy

^2.0.0-beta.2 / 1.0.2

### Describe the bug

Found while poking around replacing the built-in Markdown engine (see #2777) that when setting a replacement with `.addExtension('md', { ... })`, [as exemplified in the docs](https://www.11ty.dev/docs/languages/custom/#overriding-a-built-in-template-language), any explicit permalinks present in the frontmatter of `.md` sources will get processed with the replacement engine, as observed a while back (#1019).

This is a problem because most Markdown processors (including `marked` and `remark`) will wrap a plain string such as `/path/to/output/` in `<p>` tags, thus breaking the output. I'm not sure if the default engine skips the Markdown step or whether `markdown-it` simply doesn't wrap a plain string, but the issue only manifests when changing to a different engine.

### Reproduction steps

Eleventy config:

```js
const marked = require('marked');
module.exports = config => {
	config.addExtension("md", {
		compile: inputContent => data => marked.parse(inputContent)
	});
};
```

Content file:

```md
---
title: Hello world
permalink: hello-world/
---

## Heading

Paragraph.
```

Running Eleventy outputs the HTML to `_site / <p>hello-world / < / p>`.

### Expected behavior

If permalinks passing through the Markdown engine is inevitable, it would be great to get enough arguments to the `compile()` function for authors to be able to discern when to skip Markdown processing.

danburzo:
The workaround is to use the `compileOptions.permalink` option, which can be set to `"raw"` in case all permalinks in the project are static. 

For dynamic permalinks, the solution depends on a fix to #2777, as they are not currently pre-processed with Liquid/Nunjucks when the Markdown template engine is swapped.

zachleat:
I think a very good case can be made to make the default `raw` and encourage permalink callback functions in custom syntaxes by default!

zachleat:
Shipping with 3.0.0-alpha.13

zachleat:
Docs deep link https://www.11ty.dev/docs/languages/custom/#compileoptions.permalink-to-override-permalink-compilation

danburzo:
Cool, thanks!

jnv:
[...]

Issue #2777, opened 2023-01-27
Allow custom markdown engines to access preprocessing engine


danburzo:
### Is your feature request related to a problem? Please describe.

I'm trying to replace `markdown-it` with `remark`, and the docs currently state:

> Note that overriding md opts-out of the default pre-processing by another template language of [Markdown Files](https://www.11ty.dev/docs/config/#default-template-engine-for-markdown-files).

### Describe the solution you'd like

It would be great if we had access to the Liquid/Nunjucks/etc. pre-processing to make a fully-fledged replacement. 

### Describe alternatives you've considered

Mimicking the `this.mdLib` API like [florianeckerstorfer/eleventy-plugin-remark](https://github.com/florianeckerstorfer/eleventy-plugin-remark) is currently doing.

danburzo:
For a Nunjucks-specific workaround, you can promisify the event to wait for, and obtain, a reference to the library and associated environment:

```js
const njk = new Promise(resolve => {
  config.events.on('eleventy.engine.njk', payload => resolve(payload));
});
```

zachleat:
This seems marginally related to the Render plugin, fwiw: https://www.11ty.dev/docs/plugins/render/

Also worth mentioning https://www.11ty.dev/docs/languages/custom/#overriding-a-built-in-template-language (which is not quite what you were asking for)

zachleat:
More recently I would note (for future me or other interested parties) that simple aliases of an existing markdown engine _do_ use the preprocessing engine, as #3339 notes.

Likely this would involve an addition similar to `defaultRenderer` or might look like #3351.

> By default this repo closes enhancement request issues and places them in a queue for folks to upvote. We do this to avoid a large build-up of open issues. We got this idea from the [lodash project](https://twitter.com/samselikoff/status/991395669016436736).

> Don’t forget to upvote the first comment with 👍 to register your vote! You can view [the current enhancement backlog](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+).

PR #2781, opened 2023-01-29
Make Markdown library .set() method optional

danburzo:
This is a tiny improvement in ergonomics for replacing the underlying Markdown library. 

Due to some issues with `eleventyConfig.addExtension('md')`, namely that it opts you out of pre-processing (#2777) and that it causes issues with permalinks (#2780), `eleventyConfig.setLibrary('md')` remains the best way to replace the library.

At the moment, the library is tacitly assumed to be `markdown-it`, due to usage of library-specific methods (`set()`, `render()`, `disable()`), but there's no other reason why another library wouldn't work, as long as it provides a `render(str, data)` method. 

This PR makes the `set()` method optional in the replacement library.


zachleat:
This will ship with v2.0.0 thank you!

Issue #2604, opened 2022-10-13
Windows have EMFILE: too many open files problem


ausir0726:
**Describe the bug**
When I use 11ty develope my site it's work very well.
but when I put all my archive articles and image to 11ty and build it
it's have EMFILE: too many open files error
it's total about 7000 iamges and 4000 pages (each page have 2 layer layout with pug )
( image copy done, pug complier done, but write html fail )
I tried disable addPassthroughCopy my images, and it's work fine.
seems like addPassthroughCopy and write html will reach the open files limit at same time.

It's only happen on windows ( win10 pro installed WSL 2 docker)

**To Reproduce**
`eleventy --input=build`

**Expected behavior**
build success, but fail.

**Screenshots**
![image](https://user-images.githubusercontent.com/7247064/195484621-514769cf-88f4-481f-91a9-3290268c4edd.png)

**Environment:**

- win10 pro installed WSL 2 docker
- tried any version of 11ty, 0.12 , 1.0.2 2.0.0-canary.16

**Additional context**
it's work fine on linux and mac.

ausir0726:
I'm using a little trick for fix this at moment
it's work fine

the build folder include all my archive articles about 4000 pages.
```js
const yargs = require('yargs/yargs');
const copy = require('recursive-copy');

module.exports = (eleventyConfig) => {
  eleventyConfig.setUseGitIgnore(false);
  const argv = yargs(process.argv).argv;

  if (argv.input == 'src') {
    eleventyConfig.addPassthroughCopy(`${argv.input}/images`);
    eleventyConfig.addPassthroughCopy(`${argv.input}/spotlight/**/*.jpg`);
    eleventyConfig.addPassthroughCopy(`${argv.input}/english/spotlight/**/*.jpg`);
  }
  if (argv.input == 'build') {
    copy(`${argv.input}/images`, '_site/images', { overwrite: true });
    copy(`${argv.input}/spotlight`, '_site/spotlight', { overwrite: true, filter: ['**/*.jpg'] });
    copy(`${argv.input}/english/spotlight`, '_site/english/spotlight', { overwrite: true, filter: ['**/*.jpg'] });
  }
}
```

pdehaan:
Related: https://github.com/11ty/eleventy/issues/2139
Wonder if copying the 7k images outside of 11ty (using `rsync` or `cp` or something) would handle EMFILE errors better.

ausir0726:
> 

I tried use `rsync` or `cp` it's work very well
[...]

PR #2915, opened 2023-04-25
Refactor/switch to async fs operations after init/config load

epelc:
This patch switches almost all usage of the `fs.*Sync` apis and `fs.promises.*` to using `graceful-fs` async promise versions instead. Required changing several methods and functions to be async but I believe they are all internal to the eleventy api(aka not a breaking change).

The only parts of eleventy which still use the sync apis after this are the initial config load ones.

## EMFILE Issues on windows/towards potential fix
There has been a history of EMFILE issues on windows with elventy. We've experience this in our own projects even on latest 2.0.1. These mostly stem from `addPassthroughCopy` but people have also had issues with a large number of template files. In our case we have a large number of image files(10k+).

https://github.com/11ty/eleventy/issues/2604

Although this patch does not fix it. Adding a graceful fs patch at the top of `Eleventy.js` after this patch does make large projects work but very slow on windows. Example: on windows with 16k files build takes 120 seconds instead of 12 under wsl or macOS. Previously the build would error out with EMFILE no matter what. I believe it has something to do with watchers or `recursive-copy` somehow.

## Improvements to eleventy
Anyways regardless of our own EMFILE issues this patch should improve the resiliency of eleventy since graceful-fs handles many other errors. The project already uses graceful-fs in a few places so I wouldn't think you'd oppose. I've also gone ahead and update all of the tests so they use `async/await` where necessary due to `getOutputPath()` changing.

## Performance
On wsl, linux, and macos this doesn't affect performance. On windows it makes large projects which previously failed buildable although slow(something is better than nothing).

Perf numbers on macOS m2 max:
``` 
Various plain eleventy versions:
[11ty] Copied 16549 files / Wrote 1191 files in 12.54 seconds (10.5ms each, v2.0.0-canary.9)
[11ty] Copied 16549 files / Wrote 1191 files in 11.50 seconds (9.7ms each, v2.0.1)

With: graceful-fs async all ops patch
[11ty] Copied 16549 files / Wrote 1191 files in 11.59 seconds (9.7ms each, v2.0.2-alpha.1)
```

Performance is essentially unchanged. Especially if you run multiple times this fluctuates.

### DIFF
--- a/src/Engines/TemplateEngine.js
+++ b/src/Engines/TemplateEngine.js
@@ -1,4 +1,6 @@
-const fs = require("fs");
+const fs = require("graceful-fs");
+const util = require("util");
+const fsReadFile = util.promisify(fs.readFile);
 const { TemplatePath } = require("@11ty/eleventy-utils");
 
 const TemplateConfig = require("../TemplateConfig");
@@ -155,8 +157,7 @@ class TemplateEngine {
             partialPathNoExt = TemplatePath.removeExtension(partialPathNoExt, "." + extension);
           });
 
-          return fs.promises
-            .readFile(partialFile, {
+          return fsReadFile(partialFile, {
               encoding: "utf8",
             })
             .then((content) => {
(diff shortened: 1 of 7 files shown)


zachleat:
Merged, shipping (soon-ish) with Eleventy v3.0.0-alpha.1. Thank you!

Issue #2920, opened 2023-04-26
Translations of URLs doesn't work with strict errorMode


mathieuprog:
### Operating system

Windows 11

### Eleventy

2.0.1

### Describe the bug

When setting the errorMode to strict, I have the following error when trying to use `lcoale_url` (when pages for default language are being built):

> [11ty]   Error: Localized file for URL /en/en/features/ was not found in your project. A non-localized version does exist—are you sure you meant to use the `locale_url` filter for this? You can bypass this error using the `errorMode` option in the I18N plugin (current value: "strict"). (via Template render error)

Demo: https://github.com/mathieuprog/11ty_website

When setting errorMode to "allow-fallback" it works.

### Reproduction steps

clone
npm start

zachleat:
Shipping with 3.0, thank you!

PR #2921, opened 2023-04-26
Fix URL normalization when URL is current language

mathieuprog:
Fixes https://github.com/11ty/eleventy/issues/2920

### DIFF
--- a/src/Plugins/I18nPlugin.js
+++ b/src/Plugins/I18nPlugin.js
@@ -250,6 +250,10 @@ function EleventyPlugin(eleventyConfig, opts = {}) {
         contentMaps.localeUrlsMap[url] ||
         (!url.endsWith("/") && contentMaps.localeUrlsMap[`${url}/`])
       ) {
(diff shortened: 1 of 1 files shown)


zachleat:
Sorry for the super late response here, I’m going through active PRs for the 3.0 release.

Just thinking out loud here, I think there is more to fix: 

* `{{ "/features" | locale_url }}` resolves to `/fr/features` (correct) and `/features` in `/en/` dir (incorrect)
* `{{ "/en/features" | locale_url }}` errors incorrectly (I believe the subject of this PR—per the [docs](https://www.11ty.dev/docs/plugins/i18n/#locale_url-filter) this should work)
* `{{ "/fr/features" | locale_url }}` resolves to `/fr/features` in `/fr/` and `/en/` (incorrect, this should output `/en/features` in `/en/` dir)
* `{{ "/hi/features" | locale_url }}` resolves to `/hi/features` in `/fr/` and `/en/` (this one is fine as it doesn’t match any lang in the project)

zachleat:
Shipping with 3.0, thank you!

PR #2971, opened 2023-06-22
Sync args input&output with config

vanarok:
Add pass input and output from arg to eleventyConfig


vanarok:
I use the bundler plugin and it writes to the build in _site when the arguments are --output dist

zachleat:
We’re encouraging folks to use the new `eleventyConfig.directories` feature instead, see #3244.

Thank you though!

PR #2973, opened 2023-06-23
Update README.md

Irishgeoff11:
Adding resource


zachleat:
Probably—if it makes sense! I don’t know where it might make sense on the
11ty-website repo for 11ty.dev but look around! We have plenty of other
integration and plugin links there

Issue #2741, opened 2023-01-11
Pug includes used only in other _includes trigger a re-render on --watch but aren't processed.


yhorian:
### Operating system

Windows 11

### Eleventy

v2.0.0-canary.28

### Describe the bug

If an _include file (nav.pug) that is used in another _include file (default.pug) is changed, and it's not used anywhere outside of _includes, then when Eleventy rebuilds it does not re-render the first include (nav.pug) when it updates the second (default.pug).

This means using a component layout with _includes in Pug, if the include is not directly used by something outside the directory you have to force a full render to work on the file.

### Reproduction steps

1. Add a new file: _include/test.pug
2. Add text: `h2 Include Test`
3. Add this to the top of a current layout, such as _include/default.pug using `include ./_include/test.pug`
4. Run Eleventy in dev mode. "npx eleventy --watch --serve"
5. Change _include/default.pug by adding -> `p Layout Test`. It will trigger watch as expected, adding in the new element.
6. Change _include/test.pug by adding -> `p Include Test`. This will trigger watch but the new element won't be added to the rendered HTML.

### Expected behavior

It's expected that when you trigger a re-render with --watch that the code is included!

zachleat:
`--incremental` or no?

yhorian:
Tried with both, and neither wanted to save more than once. I also ran the tests with nunjucks - worked as expected.

I ran another test using Pugs extends/blocks instead of the eleventy system. That all works 100%. Renders any length of child extends/blocks no problems.

It's just Pug and includes. It won't save more than once with --incremental, or without. Here's my test setup for the bug:

/test-page.pug
```
---
title: Testing pug inheritance
permalink: "| /test-pug/"
layout: test_layout.pug
---

h1 #{ title }
```
/_includes/test-layout.pug - parent include
```
p 1st:
    | !{ content }
p 2nd:
    include partials/test_include.pug - child include
```
/_includes/test-include.pug - page
```
p Test include here

```

If you edit and save the child include then it won't update even though it triggers a file change on watch.

If you then go to edit and save page it still won't update.

If you edit and save the parent include it will finally update the template with the current child include.

Seems that any _include/ file that uses another _include/ file won't fetch a fresh render of it in Pug using --watch unless it's refreshed itself. You can see this behaviour by adding an include for the child include to the page as well. The page will always have the freshest version while continuing to display the old version until the parent include is refreshed.

Zearin:
[...]

PR #3059, opened 2023-09-29
Register dependencies found during pug compilation

aschrab:
My main layout is written in pug, and uses an `include` file. When running `eleventy --serve` I noticed that changes to the included file would trigger rebuilds, but not show the changes. This seems to be similar to #2741.

Here I fix this by using the list of dependencies that the pug `compile` function includes on the function that it returns. With this I am seeing changes to my included file reflected immediately.

### DIFF
--- a/src/Engines/Pug.js
+++ b/src/Engines/Pug.js
@@ -35,7 +35,13 @@ class Pug extends TemplateEngine {
       options.filename = inputPath;
     }
 
-    return this.pugLib.compile(str, options);
+    const compiled = this.pugLib.compile(str, options);
+
+    if (compiled.dependencies) {
+      this.config.uses.addDependency(inputPath, compiled.dependencies);
+    }
+
+    return compiled;
   }
 }
 


zachleat:
Related #3081

Zearin:
@aschrab Since native Pug support is gone in Eleventy 3, I am working on a Pug plugin. It’s still early, but if you use any alpha releases of Eleventy 3, I would be grateful if you tried the plugin and let me know what needs work!

It’s at: [Zearin/eleventy-plugin-pug](https://github.com/Zearin/eleventy-plugin-pug). (It’s not published on NPM at this time.)

aschrab:
> @aschrab Since native Pug support is gone in Eleventy 3, I am working on a Pug plugin. It’s still early, but if you use any alpha releases of Eleventy 3, I would be grateful if you tried the plugin and let me know what needs work!

@Zearin Yeah, I'd tried that a couple times in the past and ran into some issues, but didn't really have time to try to determine if the problems were in my code or to file a proper problem report. Just tried again now and my site is still broken with the update, but that's definitely due to some of the odd things that I'm doing.

After removing the biggest hack from my site's codebase I was at least able to get builds working, and it looks like the issue that this PR was supposed to solve isn't present in your plugin. Changes that I made to included templates, extended templates, and even templates included from an extended template and all of those were immediately reflected by the dev server.

I probably won't be doing any extensive testing until I can find a fix or alternative to the removed hacky code. If I find any issues at that point I'll definitely report them once I can isolate them, hopefully with fixes.

Issue #691, opened 2019-09-10
Expose webhook during serve for use with external data


mattwaler:
**Is your feature request related to a problem? Please describe.**

There is no easy way to trigger an update / rebuild during the serve command when external data changes.

This would be extremely useful in two ways:
- Once external webhooks are configured, would trigger a rebuild/reload locally when content changes (better DX overall)
- Would allow developers to create a "staging" site on another server that is constantly rebuilding with the newest content for preview before triggering a production build

**Describe the solution you'd like**

It would be awesome if the --serve command exposed a webhook that would trigger a rebuild, similar to when actual files change. Something like `{url}/__rebuild`

**Describe alternatives you've considered**

Any alternative would be awesome if it accomplished the same goal, allowing an external webhook to trigger a rebuild.

antgel:
@mattwaler This was a while ago, but why did you close it? I've been looking for something like this for Strapi / 11ty local development and haven't found anything (maybe I've been looking in the wrong places).

antgel:
I think https://github.com/11ty/eleventy/pull/3085 and https://github.com/11ty/eleventy-dev-server/pull/70 should help with this, if anyone is reading and interested.

Issue #604, opened 2019-07-10
Programmatically triggering an 11ty rebuild 


khawkins98:
I've successfully mapped a JS object ([like this](https://www.11ty.io/docs/data-js/#example%3A-exposing-environment-variables)) and that's working smoothly. 

What I'd now like to do is for the 11ty watch to trigger when that javascript changes. 

Conceptual example:

`myData.js`
```js
module.exports = {
  numberOfTweets: somePassedObjectOfTweets.tweetCount
};
```

Then when `somePassedObjectOfTweets` changes, I'd like for the 11ty watch to trigger; sort of something like this:

`.eleventy.js`
```js
Object.observe(somePassedObjectOfTweets, function() {
  trigger11tyWatch();
});
```

Any ideas?

## My workaround

For the time being I'm just doing this:

 `.eleventy.js`
```js
const touch = require("touch");
// https://www.npmjs.com/package/touch
Object.observe(somePassedObjectOfTweets, function() {
   touch('src/site/_data/myJsDataFile.js');
});
```

chrisssycollins:
I'm confused, sorry. Do you need/want to rebuild your site when you post a tweet? 
What would cause the contents of your `myData.js` file to change?

khawkins98:
In this case I have a parallel node process running (component library: fractal) and the object of components is exposed to 11ty via data.js file, however I need 11ty to take note when that object updates. 

The data object is successfully passed on fractal update, however 11ty doesn't trigger its `watch` rebuild. 

So for now it's working by "touching" a watched folder on fractal update. 

It would be great if there was some kind of rebuild function available inside `eleventy.js`

khawkins98:
If that's still not clear, here's my commit with my working `touch` workaround: https://github.com/visual-framework/vf-demo-design-system/commit/69f404d7e35fcbff32f16b44452ec8fc101200d3

Ryuno-Ki:
I'm wondering, whether it's the right approach.
I mean, eleventy is a SSG.
I think, what you're looking for is something, which invokes a command whenever a certain event is received.
I'm pretty sure, that there is such a library.
The command would then trigger a build.

chrisssycollins:
@Ryuno-Ki I agree, @khawkins98 I feel your approach is perhaps trying to use the 11ty config as a task runner. I'd probably use gulp to watch the files then trigger an eleventy build when your files have changed.

khawkins98:
[...]

PR #3085, opened 2023-10-30
Support rebuild event from dev server

antgel:
This feature adds the ability to define a `rebuildUrl` in the config. When eleventy-dev-server is running and a POST request is made to that URL, 11ty rebuilds the site.

Rationale: I'm working on a cool 11ty / Strapi project. In the cloud, when content is updated on Strapi, Strapi can send a webhook to whichever service is hosting 11ty, to rebuild the 11ty site. However this doesn't work for local development, as discussed at https://github.com/11ty/eleventy/issues/604, https://github.com/11ty/eleventy/issues/691 and elsewhere.

There have been suggested workarounds that generally involve the 11ty user writing an(other) HTTP server to accept the request and dumping a file somewhere to trigger a rebuild. But it is clunky and we already have an HTTP server. I reused existing EventBus functionality; really most of my time was spent reading and understanding the code.

I hope this is useful; it's my first 11ty contribution so feel free to let me know how it may be improved. :)

Corresponding dev server PR: (to be filled in)

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -952,6 +952,8 @@ Arguments:
     benchmark.before();
     await this._initWatchDependencies();
     benchmark.after();
+
+    eventBus.on("eleventyDevServer.rebuild", async () => await this._watch());
   }
 
   get isEsm() {
--- a/src/EleventyServe.js
+++ b/src/EleventyServe.js
@@ -3,6 +3,7 @@ import EleventyDevServer from "@11ty/eleventy-dev-server";
 import debugUtil from "debug";
 
 import EleventyBaseError from "./EleventyBaseError.js";
+import EventBus from "./EventBus.js";
 import ConsoleLogger from "./Util/ConsoleLogger.js";
 import PathPrefixer from "./Util/PathPrefixer.js";
 import merge from "./Util/Merge.js";
@@ -164,6 +165,10 @@ class EleventyServe {
 
     // Static method `getServer` was already checked in `getServerModule`
     this._server = serverModule.getServer("eleventy-server", this.outputDir, this.options);
+
+    if ("setEventBus" in this._server) {
+      this._server.setEventBus(EventBus);
+    }
 
     this.setAliases(this._aliases);
   }


Aankhen:
> There have been suggested workarounds that generally involve the 11ty user writing an(other) HTTP server to accept the request and dumping a file somewhere to trigger a rebuild. But it is clunky and we already have an HTTP server.

That’s right, I suggested a bespoke HTTP server that uses Eleventy’s programmatic interface. My main concern is that implementing this means encouraging what is only meant to be a dev server to accept arbitrary Internet requests, increasing its responsibilities and attack surface. Given that the requirements you’ve outlined are reasonable but also atypical, it seems fair to me to place some of the burden in this scenario on the user: I would personally prefer an out-of-tree plugin, a wrapper script, or even a code snippet in the documentation to integrating it into Eleventy proper.

antgel:
Thanks for the comment, yes indeed I forgot to mention that [suggestion](https://discord.com/channels/741017160297611315/1167159034101629088) in the OP.

> My main concern is that implementing this means encouraging what is only meant to be a dev server to accept arbitrary Internet requests, increasing its responsibilities and attack surface.

Nobody is "encouraging" the server to do anything except accept requests, which is the point of a server. The requests are not "arbitrary", on the contrary have a very specific purpose, nor are they "Internet" unless one runs the dev server on a publicly-accessible IP address, which is orthogonal from this PR.

In fact this PR (locally) replicates the functionality that one may get on the public Internet when hosting 11ty on a service such as Netlify or Vercel which accept webhooks and rebuild the site. So increasing its responsibilities doesn't seem like a Bad Thing to me.

I have added https://github.com/11ty/eleventy-dev-server/pull/70/commits/1c242aa56d444e6dd5f21b53bad1e82e5461df03 which should go even further to assuage your concerns about the "attack surface".

> Given that the requirements you’ve outlined are reasonable but also atypical, it seems fair to me to place some of the burden in this scenario on the user

I don't think that these requirements are "atypical", as seen by the Issues linked in the OP and other various Internet discussions with hacky workarounds. Use of headless CMSs is [on the rise](https://trends.google.com/trends/explore?date=today%205-y&q=headless%20cms,strapi&hl=en-GB) .

Also, "some of the burden" on the user? Currently *all* of the burden is on the user to be able to develop Headless CMS / Jamstack with the same low-friction approach, and I haven't seen any other slick implementation out there. In any event, why not put the user first? We're talking about a handful of lines of code.

Aankhen:
> Nobody is "encouraging" the server to do anything except accept requests, which is the point of a server. The requests are not "arbitrary", on the contrary have a very specific purpose, nor are they "Internet" unless one runs the dev server on a publicly-accessible IP address, which is orthogonal from this PR.

Maybe I’ve misunderstood something. How does the server accept requests from Strapi if it isn’t publicly accessible? (And to be clear, I didn’t mean ‘arbitrary’ to refer to these well-specified webhook requests, but rather the ceaseless flood of requests that any publicly accessible server receives, deliberately or not.)

> In fact this PR (locally) replicates the functionality that one gets on the public Internet when hosting 11ty on a service such as Netlify or Vercel which accept webhooks and rebuild the site. So "increasing the server's responsibilities" in such a small way doesn't seem like a Bad Thing to me.

I understand what you’re saying and I think it’s a sensible and useful feature *in the abstract,* but it’s (naturally) extra code to design, maintain, and reason about the effects of, so it’s worth exploring the justifications and ramifications.

> I have added [11ty/eleventy-dev-server@1c242aa](https://github.com/11ty/eleventy-dev-server/commit/1c242aa56d444e6dd5f21b53bad1e82e5461df03) which should go even further to assuage your concerns about the "attack surface".

I’m definitely happy to see that. Thank you.

> I don't think that these requirements are "atypical", as seen by the Issues linked in the OP and other various Internet discussions with hacky workarounds. Use of headless CMSs is [on the rise](https://trends.google.com/trends/explore?date=today%205-y&q=headless%20cms,strapi&hl=en-GB) .

This is hard to quantify. I know headless CMSs have become quite popular; on the other hand, most Eleventy users I interact with don’t use headless CMSs… which could easily be caused in part by the absence of features like this. Either way, I do think better integration with headless CMSs is a good idea. I’m only examining the design of this particular feature.
 
> Also, "some of the burden" on the user? Currently _all_ of the burden is on the user to be able to develop Headless CMS / Jamstack with the same low-friction approach, and I haven't seen any other slick implementation out there.

I meant that Eleventy’s programmatic API gives you the low-level ability to trigger the rebuild. ‘Some of the burden’ is connecting that to whichever mechanism you want to use to trigger it. (I agree that it’s not a trivial task.)

> In any event, why not put the user first? We're talking about a handful of lines of code.

The user absolutely needs to come first, and a crucial aspect of that is not inadvertently creating vulnerabilities that don’t exist without Eleventy, which is why a feature that expects a dev server to be exposed to the Internet gives me pause. As I said above, if I’ve misunderstood how it works, then my concerns are irrelevant and this is unequivocally a great feature that I’m sure will be deeply appreciated.

antgel:
> > Nobody is "encouraging" the server to do anything except accept requests, which is the point of a server. The requests are not "arbitrary", on the contrary have a very specific purpose, nor are they "Internet" unless one runs the dev server on a publicly-accessible IP address, which is orthogonal from this PR.
> 
> Maybe I’ve misunderstood something. How does the server accept requests from Strapi if it isn’t publicly accessible? (And to be clear, I didn’t mean ‘arbitrary’ to refer to these well-specified webhook requests, but rather the ceaseless flood of requests that any publicly accessible server receives, deliberately or not.)

Thanks for clarifying. I think you have misunderstood the level of exposure required / expected to use this feature.

The intention is that the 11ty dev server accepts rebuild HTTP requests from a Strapi instance (or indeed any other service that wishes to signal to 11ty via HTTP to rebuild) that may be running:
1. On localhost e.g. 11ty on http://localhost:8080. More common i.e. Strapi and 11ty both running local servers on a development machine.
2. In the cloud tunnelling to the development machine via e.g. ngrok, serveo i.e. 11ty on https://some-random-url.ngrok.dev. This is less common but would be used if the development Strapi instance was not on localhost.

In either instance, there is no intention nor need to run the 11ty dev server on a publicly IP. (2) simulates this for certain situations, but either way, whether or not the Strapi dev server is accessible to the public Internet is irrelevant to the feature. Developers may already expose their dev server if they feel they wish or need to.

> > In fact this PR (locally) replicates the functionality that one gets on the public Internet when hosting 11ty on a service such as Netlify or Vercel which accept webhooks and rebuild the site. So "increasing the server's responsibilities" in such a small way doesn't seem like a Bad Thing to me.
> 
> I understand what you’re saying and I think it’s a sensible and useful feature _in the abstract,_ but it’s (naturally) extra code to design, maintain, and reason about the effects of, so it’s worth exploring the justifications and ramifications.

Fair enough.

> > I have added [11ty/eleventy-dev-server@1c242aa](https://github.com/11ty/eleventy-dev-server/commit/1c242aa56d444e6dd5f21b53bad1e82e5461df03) which should go even further to assuage your concerns about the "attack surface".
> 
> I’m definitely happy to see that. Thank you.

No problem and thank you for the feedback, it's an easy and useful addition.

> > In any event, why not put the user first? We're talking about a handful of lines of code.
> 
> The user absolutely needs to come first, and a crucial aspect of that is not inadvertently creating vulnerabilities that don’t exist without Eleventy, which is why a feature that expects a dev server to be exposed to the Internet gives me pause. As I said above, if I’ve misunderstood how it works, then my concerns are irrelevant and this is unequivocally a great feature that I’m sure will be deeply appreciated.

Hopefully the previous part has explained this. :)

antgel:
@zachleat This has been hanging around for nearly and a year and a half; I just rebased it. I hope you can review and merge soon; it makes developing with HTTP data sources e.g. Strapi much easier. The corresponding dev server PR https://github.com/11ty/eleventy-dev-server/pull/70 got three thumbs up and has similarly been in limbo; hope you can take a look soon. Thanks.

PR #3101, opened 2023-11-11
Remove FUNDING.yml

uncenter:
No need for this since the [organization-level FUNDING.yml](https://github.com/11ty/.github/blob/main/.github/FUNDING.yml) exists and is identical to this one.

Screenshot of a notice above this repository's FUNDING.yml file saying that it is overriding the organization level FUNDING.yml file

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,0 @@
-# These are supported funding model platforms
-
-open_collective: 11ty


zachleat:
Does the 11ty org funding file apply to forks?

uncenter:
I haven't tested that! Is that necessary? I think it's quite clear where a forked repository comes from and where credit is due if that's your concern.

zachleat:
Rolling with it!

PR #3102, opened 2023-11-13
Remove `indent_size` from `.editorconfig`

Zwyx:
As the author of this [Stack Overflow answer](https://stackoverflow.com/questions/35649847/objective-reasons-for-using-spaces-instead-of-tabs-for-indentation/75019495#75019495), I loved seeing the [recent commit which converted the codebase to tabs](https://github.com/11ty/eleventy/commit/358ec48f779fa34e14abef057cc1fa0c1a10aa45) :slightly_smiling_face: Well done :clap:

As explained in the SO answer, the indent size should only be a personal preference set in your code editor. So any dev can have their own preference.

The purpose of having tabs is to let the user chooses the indent size they prefer. Having `indent_size = 2` in the project's `.editorconfig` overrides the user's personal preference, which defeats this purpose.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,6 @@ root = true
 
 [*]
 indent_style = tab
-indent_size = 2
 end_of_line = lf
 insert_final_newline = false
 trim_trailing_whitespace = true


Georift:
I just want to add to this, the users personal preferences can be defined in an `.editorconfig` at any parent directory further up.

For example I have `~/code/.editorconfig` that defines my preferred `indent_size`, which is not taken into account if the repo also has `indent_size` set.

I did notice that VSCode is happy to ignore the repo's `indent_size` if you set `editor.tabSize: `. However neovim will only look at the `.editorconfig` file.

This [post originally](https://blog.danskingdom.com/Reasons-to-use-both-a-local-and-global-editorconfig-file/) put me onto this concept. But honestly I've never seen `root = false` change behavior at all.

zachleat:
Seems reasonable! I wonder if GitHub display preferences work similarly—let’s find out!

Zwyx:
Thanks Zach!

The tab display size can be adjusted in GitHub's [user preferences](https://github.com/settings/appearance).

I've done some tests, GitHub's UI seems to completely ignore `.editorconfig`:

- when displaying code, the tab length is what has been chosen in the user preferences;
- when editing code, the tab length can be adjusted in the editor, but it always defaults to 8:
image
- GitHub inserts spaces even with `indent_style = tab`;
- and `insert_final_newline` and `trim_trailing_whitespace` are also ignored.

Issue #2522, opened 2022-08-08
Change `filter` front matter key to `filterOut` for v2.x


Zearin:
**Is your feature request related to a problem? Please describe.**
When writing a `.filter( … ) ` function in JavaScript, I tend to think of it as “Filter FOR «truthy condition»”. In other words, I think in terms of what I want to _include_.

But the [`filter` front matter key](https://www.11ty.dev/docs/pagination/#filtering-values) filters based on what you want to _exclude_.

**Describe the solution you'd like**
For more human-friendliness, I recommend changing the name of this key to `filterOut` for the official v2.x release of Eleventy.

**Describe alternatives you've considered**
Optionally, you could allow Eleventy to continue support for `filter` with a deprecation warning.

(You could also keep the `filter` keyword, and flip the logic that Eleventy uses…but that is hostile to users that have become used to the current behavior.)

mayankkamboj47:
Maybe we could use ‘exclude’ instead of ‘filterOut’ if that’s not already taken

Snapstromegon:
I agree with @mayankkamboj47 to choose 'exclude' over 'filterOut'. Not only because it seems more semantic to me, but it also allows for the addition of an 'include' property which might filter what to include from the collection.

Zearin:
@zachleat Please don’t forget this before v2.0 leaves beta! 🙏

uncenter:
I'm all for adding `include` and `exclude` properties and deprecating `filter`.

Snapstromegon:
Especially @zachleat:
I'm willing to create a PR for this, but how would you expect it to work when both `include` and `exclude` are specified?

I personally see these options:

**1. Disallow**
Just throw/warn/whatever when both keys are specified

**2. Exclude over Include**
Exclude items even when they are included.

**3. Include over Exclude**
Include items even when they are excluded.

I'd personally tend to use option 1, but I'm completely open to suggestions.

uncenter:
Option 1 is the only one that makes sense to me, I'd go with that.

Zearin:
Maybe this can squeeze in before the final release of v3?

zachleat:
Ruthlessly punting this one to 4.0, sorry!

Zearin:
😭 

J/k. I'm good. 🙂

PR #3103, opened 2023-11-13
add: include and exclude filters for pagination

Snapstromegon:
This Commit adds the `include` and `exclude` attributes to the pagination plugin. This allows for more semantic filtering in a backwards compatible manner.

This commit also adds the required tests based on the existing filter tests.

TODO: Remove old `filter` for v3.0

fixes: #2522 with option 1 from [this comment](https://github.com/11ty/eleventy/issues/2522#issuecomment-1807181893)

### DIFF
--- a/src/Plugins/Pagination.js
+++ b/src/Plugins/Pagination.js
@@ -116,6 +116,28 @@ class Pagination {
   }
 
   isFiltered(value) {
+    const hasInclude = "include" in this.data.pagination;
+    const hasExclude = "exclude" in this.data.pagination;
+    if (hasInclude && hasExclude) {
+      throw new Error("Pagination cannot have both `include` and `exclude` filters.");
+    }
+    if (hasInclude) {
+      let included = this.data.pagination.include;
+      if (Array.isArray(included)) {
+        return included.indexOf(value) === -1;
+      }
+      return included !== value;
+    }
+    if (hasExclude) {
+      let excluded = this.data.pagination.exclude;
+      if (Array.isArray(excluded)) {
+        return excluded.indexOf(value) !== -1;
+      }
+      return excluded === value;
+    }
+
+    // Let's keep this code for backwards compatibility to V2.
+    // TODO remove in 3.0
     if ("filter" in this.data.pagination) {
       let filtered = this.data.pagination.filter;
       if (Array.isArray(filtered)) {
@@ -179,7 +201,11 @@ class Pagination {
       result = result.reverse();
     }
 
-    if (this.data.pagination.filter) {
+    if (
+      this.data.pagination.filter ||
+      this.data.pagination.include ||
+      this.data.pagination.exclude
+    ) {
       result = result.filter((value) => !this.isFiltered(value));
     }
 
--- a/test/PaginationTest.js
+++ b/test/PaginationTest.js
@@ -452,6 +452,49 @@ test("Page over an object (use values)", async (t) => {
   );
 });
 
+test("Page over an object (excluded, array)", async (t) => {
+  let tmpl = getNewTemplate(
+    "./test/stubs/paged/pagedobjectexcludearray.njk",
+    "./test/stubs/",
+    "./dist"
+  );
+
+  let data = await tmpl.getData();
+  let pages = await tmpl.getTemplates(data);
+
+  t.is(
+    (await pages[0].template.render(pages[0].data)).trim(),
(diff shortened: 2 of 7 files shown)


Snapstromegon:
It seems like there are some merge conflicts. I will take care of those.

src/Plugins/Pagination.js
```diff
@@ -99,6 +99,28 @@ class Pagination {
...
 	isFiltered(value) {
+		const hasInclude = "include" in this.data.pagination;
+		const hasExclude = "exclude" in this.data.pagination;
+		if (hasInclude && hasExclude) {
+			throw new Error("Pagination cannot have both `include` and `exclude` filters.");
+		}
+		if (hasInclude) {
+			let included = this.data.pagination.include;
+			if (Array.isArray(included)) {
+				return included.indexOf(value) === -1;
```
uncenter, line 110: I think using `.includes()` makes more sense in this scenario.
```suggestion
				return !included.includes(value);
```

src/Plugins/Pagination.js
```diff
@@ -99,6 +99,28 @@ class Pagination {
...
+			}
+			return included !== value;
+		}
+		if (hasExclude) {
+			let excluded = this.data.pagination.exclude;
+			if (Array.isArray(excluded)) {
+				return excluded.indexOf(value) !== -1;
```
uncenter, line 117:
```suggestion
				return excluded.includes(value);
```

src/Plugins/Pagination.js
```diff
@@ -162,7 +184,11 @@ class Pagination {
 			result = result.reverse();
 		}
 
-		if (this.data.pagination.filter) {
+		if (
+			this.data.pagination.filter ||
+			this.data.pagination.include ||
+			this.data.pagination.exclude
+		) {
 			result = result.filter((value) => !this.isFiltered(value));
```
uncenter, line 192: Is it worth renaming `isFiltered` to `isFilteredOut` or `isExcluded`? Or even inverting the return values and renaming it to `isIncluded`, since then you would have:
```diff
-			result = result.filter((value) => !this.isFiltered(value));
+			result = result.filter((value) => this.isIncluded(value));
```

Snapstromegon: Since main switched to v.3, I think it's okay to rename this function.

src/Plugins/Pagination.js
```diff
@@ -99,6 +99,28 @@ class Pagination {
...
+		}
+		if (hasExclude) {
+			let excluded = this.data.pagination.exclude;
+			if (Array.isArray(excluded)) {
+				return excluded.includes(value);
+			}
+			return excluded === value;
+		}
+
+		// Let's keep this code for backwards compatibility to V2.
```
uncenter, line 122: And we can remove this bit now right?

src/Plugins/Pagination.js
```diff
@@ -107,27 +107,27 @@ class Pagination {
...
 			if (Array.isArray(filtered)) {
-				return filtered.indexOf(value) > -1;
+				return filtered.indexOf(value) === -1;
 			}
 
-			return filtered === value;
+			return filtered !== value;
 		}
 
 		return false;
```
uncenter, line 133: And I think this default return value would be `true` now that the function is inverted?

src/Plugins/Pagination.js
Snapstromegon: I'd be happy to do so, but I think this should be communicated, since it affects the public API of this plugin. Is there a way to log warnings if someone still uses filters?

uncenter: Maybe we add a warning when someone uses `filter` that says "The filter Pagination property has been deprecated in Eleventy v3. You can replace it with `exclude` for the same effect".

src/Plugins/Pagination.js
Snapstromegon: Good catch, will update... 🤦

src/Plugins/Pagination.js
Snapstromegon: Yes, exactly. My question is, if there is a better way than throwing an error or using console.warn for this.

uncenter: I don't see anywhere else where we are throwing an error or console.warn-ing for a deprecation. Seems like those just get tagged as deprecated with JSDoc but aren't removed or anything.

Snapstromegon: Hey @zachleat, how would you want me to handle this here?
Maybe you even have an opinion on doing a breaking change.

PR #3112, opened 2023-11-24
Enable insert_final_newline for most files

uncenter:
This PR enables trailing newlines for most files, except for test stubs which are already ignored (some tests are failing with trailing newlines, will investigate), and adds missing trailing newlines. On why trailing newlines are good: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file, https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline.


zachleat:
Seems good! Mergin’

Thanks!

PR #3130, opened 2023-12-20
add: make config accessible from js data files

### LINKED ISSUE #3114: Add a method of using `getFilter` in files other than the Eleventy config
### Is your feature request related to a problem? Please describe.

Eleventy 0.11.0 added the ability to use Eleventy's built-in filters from within the configuration file using the `eleventyConfig.getFilter` function.

However, my projects can end up having a _lot_ of custom filters, shortcodes and other bits-n-bobs. To keep things tidy, I place my custom functions definitions in their own files rather than directly within the config function. 

For example, part of my configuration might look more like this: 

```js
const { myCoolFilter, myLessCoolFilter } = require("./config/filters/cool-ranking.js");

  // ...
  eleventyConfig.addFilter("myCoolFilter", myCoolFilter); 
  eleventyConfig.addFilter("myLessCoolFilter", myLessCoolFilter); 
  // ...
};
```

The filter function not being inline within the file prevent me from accessing the `eleventyConfig` and `getFilter` from within `cool-ranking.js` unless I manually pass it into the function, but this would seemingly make it difficult to use as a filter within templates. 

I've also encountered times where being able to access filters in `.11tydata.js` files would be useful, but there similarly doesn't seem to be a method of doing so, such as here: 

```js
module.exports = {
  eleventyComputed: {
    opengraphImageUrl: (data) => {
      // OpenGraph image URLs need to be absolute. It'd be really cool if
      // the `url` filter was available here!
      return `/images/opengraph/${data.page.fileSlug}.png`;
    },
  },
};

```

### Describe the solution you'd like

I'm not sure how practical or easy it would be, especially as filters like `url` and `get*CollectionItem` are at least somewhat dependent on the configuration itself, but it'd be nice if the built-in filters could be available in more contexts.

### Describe alternatives you've considered

At the moment I've resorted to reproducing the built-in features as custom functions, sometimes using the same dependencies as Eleventy itself, such as using the `@sindresorhus/slugify` package to duplicate the `slugify` filter.

This works in the short term, but there feels like there's a risk of the configurations being changed or falling out of sync over time, so it's not an ideal alternative in my mind. 

### Additional context

_No response_

Snapstromegon:
resolves #3114

### DIFF
--- a/src/TemplateDataInitialGlobalData.js
+++ b/src/TemplateDataInitialGlobalData.js
@@ -41,6 +41,7 @@ class TemplateDataInitialGlobalData {
 		// #2293 for meta[name=generator]
 		globalData.eleventy.version = semver.coerce(pkg.version).toString();
 		globalData.eleventy.generator = `Eleventy v${globalData.eleventy.version}`;
+		globalData.eleventy.config = this.config;
 
 		return globalData;
 	}


zachleat:
I don’t think we can move forward with this one—sorry. There are a bunch of downstream implications of placing the entire config in global data—a bit too risky!

PR #3141, opened 2023-12-31
add: make log filter async friendly by adding logAsync

### LINKED ISSUE #3133: Asynchronous filter returns a Promise object
### Operating system

macOS Sonoma 14.2.1

### Eleventy

2.0.1

### Describe the bug

I’m writing an async filter to fetch the data of a Sanity document based on its ID.
Logging the data in the console works as expected, but trying to render the result within a template doesn’t, as it returns an empty object (`Promise { <pending> }`).
```html
<code @text="sRef(reference)"></code>
```
```js
module.exports = eleventyConfig => {
  eleventyConfig.addJavaScriptFunction("sRef", async reference => {
    try {
      const response = await fetch(`https://${process.env.SANITY_STUDIO_PROJECT_ID}.api.sanity.io/v${process.env.SANITY_STUDIO_API_VERSION}/data/query/${process.env.SANITY_STUDIO_DATASET}`, {
        method: "POST",
        headers: {
          "Content-Type": "application/json",
          "Authorization": `Bearer ${process.env.SANITY_STUDIO_API_TOKEN}`
        },
        body: JSON.stringify({
          query: `*[_id == $referenceId][0]`,
          params: {
            referenceId: reference
          }
        })
      });
      if (!response.ok) {
        throw new Error(`Error fetching data: ${response.statusText}`);
      }
      const data = await response.json();
      return data.result;
    } catch (error) {
      console.error(`Error fetching data from Sanity: ${error.message}`);
      return null;
    }
  })
}
```
This is a WebC template but I don’t think it’s related to that.

Snapstromegon:
Right now the `log` filter logs the passed in values directly. This isn't always helpful when working with async values e.g. in webc components. This change adds the `logAsync` filter to also take promises as any value and awaits them, so you always get some useful output.

It could be possible to just eagerly return the value in the sync filter and then use a `.then` call to log the values, but that would leave room for unexpected data mutation, so I didn't implement it that way for now.

resolves #3133

### DIFF
--- a/src/defaultConfig.js
+++ b/src/defaultConfig.js
@@ -73,6 +73,11 @@ export default function (config) {
 		return input;
 	});
 
+	config.addFilter("logAsync", async (input, ...messages) => {
+		console.log(await input, ...(await Promise.all(messages)));
+		return input;
+	});
+
 	config.addFilter("getCollectionItemIndex", function (collection, pageOverride) {
 		return getCollectionItemIndex.call(this, collection, pageOverride);
 	});


zachleat:
Can we test for promises in `log` and `await` instead? It might be preferable to make `log` async friendly in 3.0

Snapstromegon:
If we make the log filter itself async, it would no longer be available in languages that don't support async filters (like handlebars).

I've just seen that addFilter only adds the filters to Liquid, JavaScriptFunctions and Nunjucks filters, so we could make it indeed async for everything.

zachleat:
Ah, sorry—I meant (kind of) like this

```js
config.addFilter("log", (...messages) => {
  // if any of messages are a promise, do promise things
  if(messages.find(entry => /* promise test */)) {
    // out of order, but that’s preferable to errors in non-async
    Promise.all(messages).then(msgs => console.log(...msgs));
  } else {
    console.log(...messages);
  }
});
```

some prior art (not exactly): https://github.com/11ty/eleventy/blob/d77d7fad6dfcbb872f0487345c2189981c509bcc/src/UserConfig.js#L272

Snapstromegon:
Ahh okay, I see what you mean and I will adapt it.

Snapstromegon:
Thinking about it, we could even return a promise if any of the inputs is async. That would also fix the ordering issue for async cases or am I missing something here?

Snapstromegon:
@zachleat are we intending to support only native Promises here, or also all "thenables" (so e.g. Bluebird promises and similar custom promise implementations)?
I think they would be detectable by testing if something has a `.then()` method, but that seems flaky too.

zachleat:
native promises only is fine. And I like the promise return!

Snapstromegon:
@zachleat Good catch.

Since the removal of some of the built-in templating languages, is there a reason to not just add it via `addAsyncFilter`?

PR #3153, opened 2024-01-09
docs: Add a sprinkle of JSDoc throughout

Zearin:
@zachleat Here’s a sample of JSDoc I added.

I tried to be as terse as possible. The codebase is large, so I wanted to show you a sample of what it would look like per your request (from Mastodon).


Zearin:
I’ll make a better attempt at this later.

PR #3165, opened 2024-01-13
refactor: Prefer `const` where appropriate

Zearin:
I just used eslint’s `--fix` option for a single rule, `prefer-const`.

**Rationale:**
From the [`prefer-const`](https://eslint.org/docs/latest/rules/prefer-const) page:

> If a variable is never reassigned, using the `const` declaration is better.
>
> `const` declaration tells readers, “this variable is never reassigned,” reducing cognitive load and improving maintainability.

### DIFF
--- a/src/Benchmark.js
+++ b/src/Benchmark.js
@@ -34,7 +34,7 @@ class Benchmark {
 			throw new Error("You called Benchmark after() without a before().");
 		}
 
-		let before = this.beforeTimers.pop();
+		const before = this.beforeTimers.pop();
 		if (!this.beforeTimers.length) {
 			this.timeSpent += this.getNewTimestamp() - before;
 		}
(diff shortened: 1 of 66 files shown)


Zearin:
(Also, I dunno if the engine does any sort of optimization for `const` variables, but it’s possible…I guess? 🤷 )

uncenter:
Side note: Eleventy doesn't use conventional commits so this doesn't really matter, but this PR technically would be a `refactor:`, not a `chore:`.

Zearin:
@uncenter Thanks! I have been trying to use conventional commits on my own just in case they could come in handy, but never used them on a team or anything. I appreciate the lesson! 🙏🏽 

Edited the title of the PR. (I don't _think_ I can reword the commit without force-pushing, though…)

Zearin:
@zachleat Rebased! Can you take a look when you have the chance?

Zearin (commented):
@zachleat I know you were OOO for a while. When you’re back, can you take a look when you have a moment?

Zearin:
I’ll try this again later. :P

zachleat:
Apologies—please check in before starting a large change like this! I would have cautioned—as the code suggests I prefer `let`.

Issue #1541, opened 2020-12-04
global data is not passed to liquidjs as globals


binyamin:
**Describe the bug**
With `{% render layout.html %}`, you can only access a variable if it's defined in the included template/partial or passed down to liquid via `liquidOptions.globals`. 11ty refers to "global data", which is undefined within such partials.

**To Reproduce**
See [example repo](https://github.com/binyamin/11ty-Test-Tube/tree/11ty-1541)
If you don't feel like running the example repo:
1. In a new 11ty project, reference a global data key from within an include/partial.
2. In a markdown file, `{% render %}` the include/partial.
3. Run `npx eleventy`

**Expected behavior**
I expect that data which is the same for every page (eg. `pkg`, `_data/**/*`) is passed down to liquid, whereas local variables are not.

**Environment:**
- OS and Version: Xubuntu 20.04
- Eleventy Version: 1.0.0-beta.0+sha:9042341 :wink:

**Additional context**
Related: https://github.com/harttle/liquidjs/issues/185

HugoGiraudel:
I’m not a Liquid expert, so don’t quote me on this, but I believe this is intended. The documentation on [render](https://shopify.dev/docs/themes/liquid/reference/tags/theme-tags#render) states:

> When a snippet is rendered, the code inside it does not automatically have access to the variables assigned using variable tags within the snippet's parent template. Similarly, variables assigned within the snippet can't be accessed by the code outside of the snippet. This encapsulation increases performance and helps make theme code easier to understand and maintain.

My understanding of this is that by definition, `render` does not pass through its current scope to the partial (which is overall a good thing, as they explain). Any variable that requires reading needs to be passed as an argument to the `render` call. In your case:

```liquid
{% render "foo.html", bar: bar %}
```

binyamin:
@HugoGiraudel thanks, I've seen that in the docs. I guess this is really a question of whether _eleventy's_ global data should be passed down as _liquid's_ global data.

cyberwombat:
It appears that currently globals are not available in the partials added through `render`. However the Shopify docs explicity say that it should:
_Global objects don't need to be passed down. They are accessible from all files._

sentience:
Per [the LiquidJS docs](https://liquidjs.com/tags/render.html#Passing-Variables) on passing variables to `render`ed partials,

> [globals](https://liquidjs.com/api/interfaces/liquid_options_.liquidoptions.html#Optional-globals) don’t need to be passed down. They are accessible from all files.

[...]

Issue #3172, opened 2024-01-22
Can't get access to data inside liquid file which is triggered by the 'render'


shkarinn:
### Operating system

Windows 11

### Eleventy

2.0.1

### Describe the bug

Hello.
I can't get access to data inside liquid file which is triggered by the 'render'. But depricated tag 'include' get access to data.
If I use old tag include
{% include 'header/section' %}
my data inside include file is working

For example I can get all data from 11ty global inside header/section.liquid file
{{ collections.all[0].template.templateData.globalData | debug }}

But if use new tag
{% render 'header/section' %}
I got undefined <pre>undefined</pre>

I can get access to data only outside file section.liquid
{% render 'header/section' for contact.buttons as item %}

liquid have section tag, but js version not.
liquidjs have only render, include, layout tags.
But include is Deprecated and working good.
Render tag have problem.
Maybe this problem inside 11ty...

### Reproduction steps

4. See an error

vrugtehagel:
Possibly a duplicate of https://github.com/11ty/eleventy/issues/2453

Issue #2453, opened 2022-06-23
(Liquid) Unable to access 11ty globals like page.X inside a partial rendered via the {% render %} tag


AleksandrHovhannisyan:
**Describe the bug**

Per the Liquidjs docs, [the {% include %} tag is deprecated](https://liquidjs.com/tags/include.html) and it's recommended that we use [the {% render %} tag](https://liquidjs.com/tags/render.html) instead. See also: https://github.com/11ty/eleventy/issues/2411.

While it's currently possible to use the `{% render %}` tag in 11ty, it's not possible to access globals or global data from within the rendered template. [Per the Liquid docs on render](https://liquidjs.com/tags/render.html#Render-a-Template):

> When a partial template is rendered, the code inside it can’t access its parent’s variables

However, the section on [passing in arguments to the rendered template](https://liquidjs.com/tags/render.html#Passing-Variables) clarifies that globals should be accessible:

> [globals](https://liquidjs.com/api/interfaces/liquid_options_.liquidoptions.html#Optional-globals) don’t need to be passed down. They are accessible from all files.

The `LiquidOptions` interface does have a `globals` config here: https://github.com/harttle/liquidjs/blob/master/src/liquid-options.ts#L72. And 11ty supports configuring Liquid options via `eleventyConfig.setLiquidOptions`.

Unfortunately, there is no way to pass along 11ty global data and the page variable because these are not accessible from the config as far as I can tell. How do I proceed?

**To Reproduce**

1. Clone this minimal sandbox branch: https://github.com/AleksandrHovhannisyan/11ty-sandbox/tree/render-tag
2. Run `yarn` to install 11ty, the only dependency.
3. Run `yarn serve` to start the server.
4. Observe that the argument passed into the `render`ed partial is accessible within the partial, as expected.
5. Observe that the `page` variable is not accessible within the partial (unexpected).

Code from the demo:

**eleventy.js**

```js
const TEMPLATE_ENGINE = 'liquid';

module.exports = (eleventyConfig) => {
  return {
    dir: {
        input: 'src',
        output: '_site',
        includes: '_includes',
        layout: '_layouts'
    },
    dataTemplateEngine: TEMPLATE_ENGINE,
    markdownTemplateEngine: TEMPLATE_ENGINE,
    htmlTemplateEngine: TEMPLATE_ENGINE,
    templateFormats: ['html', 'md', TEMPLATE_ENGINE],
  };
};
```

**src/_includes/partial.html**

```liquid
[...]

PR #3212, opened 2024-02-23
Allow accessing globals from rendered liquid partials

vrugtehagel:
This is a PR that fixes https://github.com/11ty/eleventy/issues/1541, and can probably be extended to include https://github.com/11ty/eleventy/issues/3172 and https://github.com/11ty/eleventy/issues/2453 as well.

In a nutshell, the problem is that liquid partials rendered with the `{% render %}` tag are encapsulated to the point they do not have access to global data configured through `addGlobalData`, global data files, directory- and page-specific data, and Eleventy-provided globals such as `page`. The only global data they have access to is that configured through LiquidJS's [`globals`](https://liquidjs.com/tutorials/options.html#globals) option (passed to `.setLiquidOptions()`).

This PR does not expose all of the data mentioned above to `{% render %}`-style partials (though can be extended to do so), but instead only exposes data configured through `addGlobalData` and global data files. LiquidJS's `globals` option has the lowest priority in this data cascade (which is consistent with the `{% include %}` equivalent).

A bit about the PR's contents; I've added a `.needsGlobals()` method to `TemplateEngine` instances. This method returns a boolean indicating whether it wants to receive all global data before rendering. Receiving the globals is not generally necessary, but as mentioned before, LiquidJS requires all global data be passed to its `globals` option in order to make it available in partials renderd with `{% render %}`. As such, by default, this `.needsGlobals()` method returns `false`; the `Liquid` engine then overrides this method with one returning `true`. Then, at the time the data is first retrieved, the necessary global data is passed to the engines that need it (for now, only the Liquid engine) using an `addGlobals` method (one the Nunjucks engine already possesses).

I also added two tests; if a test with an actual data file is desired, I'd like some pointers on how to accomplish that neatly (I didn't see other tests doing something like that).

### DIFF
--- a/src/Engines/Liquid.js
+++ b/src/Engines/Liquid.js
@@ -25,7 +25,6 @@ class Liquid extends TemplateEngine {
 		this.setLibrary(this.config.libraryOverrides.liquid);
 
 		this.argLexer = moo.compile(Liquid.argumentLexerOptions);
-
 		this.cacheable = true;
 	}
 
@@ -293,6 +292,7 @@ class Liquid extends TemplateEngine {
 
 		return async function (data) {
 			let tmpl = await tmplReady;
+
 			return engine.render(tmpl, data, options);
 		};
 	}
--- a/src/Engines/Liquid.js
+++ b/src/Engines/Liquid.js
@@ -19,7 +19,7 @@ class Liquid extends TemplateEngine {
 	constructor(name, eleventyConfig) {
 		super(name, eleventyConfig);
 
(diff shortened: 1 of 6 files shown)


vrugtehagel:
This PR has been open for a while without activity - @zachleat is this something you're interested in? I can resolve the conflicts but prefer not to do work if it'll be rejected in the end anyway 😉 

If you have suggestions for other ways to resolve the original issue, I'm happy to try those routes as well!

Issue #3206, opened 2024-02-19
Strict variables and custom filters in includes


vrugtehagel:
### Operating system

macOS Sonoma 14.2.1

### Eleventy

2.0.1

### Describe the bug

Eleventy fails to build if a custom tag is defined and used within a liquid include (if referenced with `{% render '…' %}`) while the `strictVariables` setting is switched on.

### Reproduction steps

Here's an example setup. The folder structure is just that of a basic Eleventy project, with an include:
```
project/
├── _includes/
│   └── hello.liquid
├── index.liquid
├── package.json
└── .eleventy.js
```
The page to render, `index.liquid`, might look like
```liquid
---
title: Hello!
---
{% render 'hello', name: 'Max' %}
```
The include, `hello.liquid`, is defined as
```liquid
<div>
    Hello, {{ name | make_it_foo }}!
</div>
```
We are using a custom `make_it_foo` filter that turns anything into the string "foo". Of course, this is defined in our `.eleventy.js` config file:
```js
module.exports = function(config) {
    config.setLiquidOptions({ strictVariables: true })

    config.addFilter('make_it_foo', function() {
        return 'foo'
    })
}
```
Note that we also have the `strictVariables` option switched on.

### Expected behavior

I expect the build to succeed.

The build instead fails, with the following error:
```
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble rendering liquid template ./index.liquid (via TemplateContentRenderError)
[11ty] 2. undefined variable: page, file:/Users/koen/Desktop/11ty-filter-bug/_includes/hello.liquid, line:2, col:9 (via RenderError)
[11ty] 3. undefined variable: page (via InternalUndefinedVariableError)
[11ty] 
[11ty] Original error stack trace: InternalUndefinedVariableError: undefined variable: page
[11ty]     at Context._getFromScope (/Users/koen/Desktop/11ty-filter-bug/node_modules/liquidjs/dist/liquid.node.cjs.js:2487:23)
[11ty]     at _getFromScope.next (<anonymous>)
[11ty]     at toValueSync (/Users/koen/Desktop/11ty-filter-bug/node_modules/liquidjs/dist/liquid.node.cjs.js:504:32)
[11ty]     at toValueSync (/Users/koen/Desktop/11ty-filter-bug/node_modules/liquidjs/dist/liquid.node.cjs.js:510:25)
[...]

PR #3213, opened 2024-02-23
Allow custom filters in Liquid partials in strictVariables mode

vrugtehagel:
This PR fixes https://github.com/11ty/eleventy/issues/3206.

The issue here is that in partials rendered using `{% render %}`, the `page` and `eleventy` variables do not exist, since the `{% render %}` tag is introducing a "new" scope that does not inherit the variables from the page it was called from. In `strictVariables` mode (this is an option you can pass to `.setLiquidOptions()`), this causes an error when you use any custom filter in such a partial, because Eleventy is trying to retrieve the value of the `page` and `eleventy` globals for use in the filter's definition.

This PR essentially turns the properties `page` and `eleventy` into getters; meaning that they are not retrieved until the user specifically uses them. This prevents filters that do not use these properties from causing errors in `{% render %}`-style partials.

My other PR, https://github.com/11ty/eleventy/pull/3212, is related, and if the `page` and `eleventy` variables are intended to be accessible from `{% render %}` partials, then this PR may be dismissed.

### DIFF
--- a/src/Engines/Liquid.js
+++ b/src/Engines/Liquid.js
@@ -58,8 +58,14 @@ class Liquid extends TemplateEngine {
 	static wrapFilter(fn) {
 		return function (...args) {
 			if (this.context && "get" in this.context) {
-				this.page = this.context.get(["page"]);
-				this.eleventy = this.context.get(["eleventy"]);
+				const exposedProperties = ["page", "eleventy"];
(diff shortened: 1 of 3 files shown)


zachleat:
Shipping with 3.0.0-alpha.6. Thank you!

Issue #3159, opened 2024-01-10
Getting to know Eleventy 3 codebase


Zearin:
@zachleat I’ve been trying to become familiar with the Eleventy codebase, and there are a number of things I’ve wondered about the code’s structure. 

Since v3 is on the horizon, I thought I’d try and aggregate some questions, so the answers may hopefully benefit others as well.

> **Note:** Also, I want to say I’m asking these questions with a great deal of respect and admiration for this fantastic project. 
> In the past, sometimes my words were received as confrontational (in other contexts, not here). Invariably I was blind to the danger until after I’d offended someone. I try hard to avoid this, and I usually succeed. 
> But I wanted to make sure I am clear that here I’m only interested in learning from a project I have loved for years. :) 

- **Errors.** There is an `Errors` folder, but also several Error subclasses sprinkled throughout many files. 
	- Why are Errors defined using two different practices? (Centralized vs. alongside code that throws them)
	- After all the specialized Error subclasses defined, there are many instances of `throw new Error(...)`. Is this because they are less likely to occur?
- **Template\*.js files.**  The `src/Template\*.js` files outnumber even the `src/Eleventy\*.js` files. Is there a reason these aren’t grouped into a directory?
- **Getters and Setters.** There is a curious mix of `getThing()` and `get thing()` throughout. (And likewise for setters.) Is this just historical legacy from before getters and setters were supported syntax? Are there instances where you used the `getThing()` style because there would be a problem using a `get thing()`?
  - **Example:** In the `TemplateWriter` class, I even found the following snippet:
	```js
	setEleventyFiles(eleventyFiles) {
	        this.eleventyFiles = eleventyFiles;
	    }


	    set eleventyFiles(eleventyFiles) {
	        this._eleventyFiles = eleventyFiles;
	    }
	``` 
- **Control Flow.**  Probably the most important question here: What is the control flow of the program?  
  - I’m not talking about the [Order of Operations](https://www.11ty.dev/docs/advanced-order/) page in the docs (although it is helpful!).  I mean, how does execution pass from class to class during a typical build?

Snapstromegon:
Hi @Zearin, nice to see someone trying to get familiar with the codebase and maybe even I can learn something from this issue.

I will try to respond to the best of my abilities:
[...]

PR #3227, opened 2024-03-22
refactor: Group like modules into appropriate folders

Zearin:
I really wanted to get to know the Eleventy codebase better (see #3159). 

At first I just started looking around, but I kept having to spend mental energy reminding myself “this module is related to that one over there in that folder,” and so on.

I originally started moving files around simply in pursuit of exploring the codebase—with more of the related files grouped into directories for my own mental benefit while exploring.

But since Eleventy v3 is still in alpha, I thought that perhaps it might be an opportune time to turn my exploration into a Pull Request.

@zachleat: I am self-taught, and not as disciplined as I sometimes wish—but I do **love** this project! ❤️ If I naïvely changed something that offends your sensibilities, please know that that is the _last_ thing I want. But, on the off chance you find this helpful, then I have no way of knowing unless I tried. Thanks for Eleventy! 🙏🏽


zachleat:
Yeah, this is great! Thank you!

Zearin:
Wow. Your acceptance means a lot to me. Thank you so very much! 🙇

Issue #3230, opened 2024-03-25
Update CLI `--help` text


w0whitaker:
Run `npx @11ty/eleventy --help` from the command line and the resulting list includes the following:
```
--dryrun
       Don’t write any files. Useful with `DEBUG=Eleventy* npx eleventy`
```

It's a very minor thing, but I believe it should be `DEBUG=Eleventy* npx @11ty/eleventy` to be consistent with the documentation.

I'd be more than happy to work on this soon as I figure out where the text is produced.

baptiste-roullin:
The text used in the help [is here](https://github.com/11ty/eleventy/blob/8f95fac1c4220d2b19dbc959861cd3d3e3650a21/src/Eleventy.js#L679).

w0whitaker:
> The text used in the help [is here](https://github.com/11ty/eleventy/blob/8f95fac1c4220d2b19dbc959861cd3d3e3650a21/src/Eleventy.js#L679).

Thank you, I've submitted a PR with new wording.

zachleat:
Awesome, thank you!

If someone else wanted to contribute changing that command when we know we’re on Windows that would be a further improvement!

Example on https://www.11ty.dev/docs/debugging/

zachleat:
Shipped with 3.0.0-alpha.6

PR #3231, opened 2024-03-27
Fixes a small typo in the output of the `--help` command.

w0whitaker:
This closes [3230](https://github.com/11ty/eleventy/issues/3230). Changed the wording for the `--dryrun` help text to use the correct command for DEBUG mode. The new text also makes it a little more general in case someone has a different setup or expectations.


zachleat:
Great, thank you!

PR #3243, opened 2024-04-05
Small simplifications

Zearin:
These are an assortment of small tweaks. 

My goal was to make these bits of code easier to understand at a glance. There’s plenty of subjectivity in that, to be sure.

Some of these make use of Javascript features that were not available a few years ago. (Both syntax and APIs of built-in objects.) Hopefully, I applied them in a way that makes it easier to understand each block of code. Since Eleventy v3 is still in alpha, I thought this would be a good time to introduce them, since they are supported pretty much everywhere and a major version increment doesn’t have to worry about breaking compatibility.

### DIFF
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -774,15 +774,19 @@ Arguments:
 
 	shouldTriggerConfigReset(changedFiles) {
 		let configFilePaths = new Set(this.eleventyConfig.getLocalProjectConfigFiles());
-		if (changedFiles.intersection(configFilePaths).size > 0) {
-			return true;
-		}
-
-		for (const configFilePath of configFilePaths) {
-			// Any dependencies of the config file changed
-			let configFileDependencies = new Set(this.watchTargets.getDependenciesOf(configFilePath));
-			if (changedFiles.intersection(configFileDependencies).size > 0) {
+		//	Has the config file itself changed?
+		for (let filePath of changedFiles) {
+			if (configFilePaths.has(filePath)) {
 				return true;
+			}
+		}
+		//	Has the config file’s depdendencies changed?
+		for (const configFilePath of configFilePaths) {
+			let configFileDependencies = new Set(this.watchTargets.getDependenciesOf(configFilePath));
+			for (let filePath of changedFiles) {
+				if (configFileDependencies.has(filePath)) {
+					return true;
+				}
 			}
 		}
 
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -390,14 +390,14 @@ class Eleventy {
 		// Restore from cache
 		if (this._privateCaches.has(key)) {
 			let c = this._privateCaches.get(key);
-			for (let cacheKey in c) {
-				inst[cacheKey] = c[cacheKey];
+			for (let [cacheKey, cacheValue] of Object.entries(c)) {
+				inst[cacheKey] = cacheValue;
 			}
 		} else {
 			// Set cache
 			let c = {};
-			for (let cacheKey of inst.caches || []) {
-				c[cacheKey] = inst[cacheKey];
+			for (let [cacheKey, cacheValue] of Object.entries(inst.caches || [])) {
+				c[cacheKey] = cacheValue;
 			}
 			this._privateCaches.set(key, c);
 		}
--- a/src/Eleventy.js
+++ b/src/Eleventy.js
@@ -774,20 +774,15 @@ Arguments:
 
 	shouldTriggerConfigReset(changedFiles) {
 		let configFilePaths = new Set(this.eleventyConfig.getLocalProjectConfigFiles());
-		for (let filePath of changedFiles) {
-			if (configFilePaths.has(filePath)) {
-				return true;
-			}
+		if (changedFiles.intersection(configFilePaths).size > 0) {
+			return true;
 		}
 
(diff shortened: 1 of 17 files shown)


src/Eleventy.js
```diff
@@ -778,20 +778,15 @@ Arguments:
 
 	shouldTriggerConfigReset(changedFiles) {
 		let configFilePaths = new Set(this.eleventyConfig.getLocalProjectConfigFiles());
-		for (let filePath of changedFiles) {
-			if (configFilePaths.has(filePath)) {
-				return true;
-			}
+		if (changedFiles.intersection(configFilePaths).size > 0) {
```
uncenter, line 781: Forgive my ignorance but is this really supported in Node yet?

The following works in my Chrome browser console, but not in the latest version of Node.js (v21.7.2).

```
new Set([1, 2, 3]).intersection(new Set([2, 3, 4]))
```

Screenshot 2024-04-05 at 17 05 52 (Ghostty)
Screenshot 2024-04-05 at 17 06 47 (Arc)

zachleat: node.green has this only in Nightlies so far. We need all code to be Node 18+ for the 3.0 release.

Zearin: Aw, dang! **Still** no set operations in JS? 🤦‍♂️ 

Come on, Node.js devs…


(I’ll have this fixed shortly!)

src/EleventyExtensionMap.js
```diff
@@ -96,12 +89,11 @@ class EleventyExtensionMap {
...
-		for (let extension in this.extensionToKeyMap) {
-			if (path.endsWith(`.${extension}`)) {
-				extensions.add(extension);
-			}
-		}
+		let extensions = new Set(
+			Array.from(Object.keys(this.extensionToKeyMap)).filter((extension) =>
+				path.endsWith(`.${extension}`),
+			),
+		);
```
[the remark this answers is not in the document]
Zearin: `Object.keys()` returns an Iterator, not an array, so it needs to array-ified first. But I can just change it to `[ … ]` with spread syntax.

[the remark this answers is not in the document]
uncenter: > `Object.keys()` returns an Iterator, not an array, so it needs to array-ified first. But I can just change it to `[ … ]` with spread syntax.

Are you sure? MDN says "`Object.keys()` returns an array whose elements are strings"...
image

Zearin:
Updated based on your feedback. 

Whatcha think?

src/EleventyExtensionMap.js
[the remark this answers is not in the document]
uncenter: This still seems relevant, I think my comment got missed.

src/EleventyExtensionMap.js
uncenter, line 1: See https://github.com/11ty/eleventy/pull/3243#discussion_r1558398155.

Zearin: Fixed in c69fe27.

src/EleventyExtensionMap.js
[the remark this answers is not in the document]
Zearin: Fixed in c69fe27.

src/Eleventy.js
```diff
@@ -769,16 +769,15 @@ Arguments:
...
 	shouldTriggerConfigReset(changedFiles) {
 		let configFilePaths = new Set(this.eleventyConfig.getLocalProjectConfigFiles());
+		//	Has the config file itself changed?
 		for (let filePath of changedFiles) {
 			if (configFilePaths.has(filePath)) {
 				return true;
 			}
 		}
-
+		//	Has the config file’s depdendencies changed?
```
uncenter, line 778:
```suggestion
		//	Has the config file’s dependencies changed?
```

src/Eleventy.js
```diff
@@ -794,12 +793,10 @@ Arguments:
 		if (!activeQueue.length) {
 			return false;
 		}
-
-		return this.shouldTriggerConfigReset(
-			activeQueue.map((path) => {
-				return PathNormalizer.normalizeSeperator(TemplatePath.addLeadingDotSlash(path));
-			}),
+		activeQueue = new Set(
+			activeQueue.map(TemplatePath.addLeadingDotSlash).map(PathNormalizer.normalizeSeperator),
```
uncenter, line 797: I know it looks nice but wouldn't it be better to only map over it once?

src/EleventyExtensionMap.js
```diff
@@ -167,18 +157,13 @@ class EleventyExtensionMap {
 	}
 
 	hasExtension(key) {
-		for (var extension in this.extensionToKeyMap) {
-			if (this.extensionToKeyMap[extension] === key) {
-				return true;
-			}
-		}
-		return false;
+		return [...Object.values(this.extensionToKeyMap)].includes(key);
```
uncenter, line 160:
```suggestion
		return Object.values(this.extensionToKeyMap).includes(key);
```

src/Eleventy.js
Zearin: Well after prettier reformatted what I wrote (grumble), I'm not even sure it looks better. I'll change back.

Zearin:
I messed up on a rebase. 🤦‍♂️ I’ll redo this in a new PR.

PR #3272, opened 2024-05-01
Write templates synchronously

### LINKED ISSUE #2627: EMFILE: too many open files when attempting to render a large (170000) number of pages (Eleventy 2.0.0-canary.16)

### LINKED ISSUE #3271: [Performance] Writing templates is slow because of unbounded concurrency
### Operating system

Windows 10

### Eleventy

3.0.0-alpha.9

### Describe the bug

This use of Promises:

https://github.com/11ty/eleventy/blob/9c3a7619a429f26cc2b449d65209be6d6a23828e/src/Template.js#L751

Means that almost all templates are written at the same time to the disk, which dramatically slows down individual writes. I added a `console.time` around the write to show this and recorded [the output](https://kaagaz.shivjm.in/related-roughy.log). You can see that some files take a full 15 seconds to write, and the total time for ‘Template Write’ is 15,885ms or nearly 16 seconds with 926 files.

For the sake of testing, I changed just that line to use `fs.writeFileSync` instead of `await writeFile`. [Here’s the output.](https://kaagaz.shivjm.in/loving-hound.log) The time for ‘Template Write’ dropped to 232ms, with each individual file taking under a millisecond.

### Reproduction steps

1. Create multiple template files.
2. Run Eleventy.

### Expected behavior

Files are written sequentially or with bounded concurrency.

### Reproduction URL

_No response_

### Screenshots

_No response_

shivjm:
I’m not sure how to implement a queue to resolve #3271 but here’s my proposed workaround. This does block everything while writing a template (hopefully a small window).

Closes #3271
Closes #2627

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -25,7 +25,6 @@ import EleventyBaseError from "./Errors/EleventyBaseError.js";
 import ReservedData from "./Util/ReservedData.js";
 
 const { set: lodashSet, get: lodashGet } = lodash;
-const writeFile = util.promisify(fs.writeFile);
 const fsStat = util.promisify(fs.stat);
 
 const debug = debugUtil("Eleventy:Template");
@@ -748,7 +747,9 @@ class Template extends TemplateContent {
 		let templateBenchmark = this.bench.get("Template Write");
 		templateBenchmark.before();
 
-		await writeFile(outputPath, finalContent);
+		// Note: This deliberately uses the synchronous version to avoid
+		// unbounded concurrency: https://github.com/11ty/eleventy/issues/3271
+		fs.writeFileSync(outputPath, finalContent);
 
 		templateBenchmark.after();
 		this.writeCount++;


shivjm:
@uncenter pointed out that the Fetch and Image plugins use [the p-queue library](https://www.npmjs.com/package/p-queue). I think I’ll add a commit to this PR using p-queue; if it’s not desirable, I can back it out.

zachleat:
I’m okay with shipping this on in 3.0.0-alpha.11.

Preliminary benchmarking suggests that the performance cost is about the same here:

```
---------------------------------------------------------
Eleventy Benchmark (Node v20.12.2, 10000 templates each)
---------------------------------------------------------
Eleventy 3.0.0-alpha.10                                        
---------------------------------------------------------
.md: ... 3 runs:              
* Median: 7.92 seconds
* Median per template: 792 µs

---------------------------------------------------------
Eleventy 3.0.0-alpha.11                                        
---------------------------------------------------------
.md: ... 3 runs:              
* Median: 8 seconds (1%)
* Median per template: 800 µs (1%)
```

zachleat:
We can continue some of the other async queuing discussion on #3277.

PR #3275, opened 2024-05-03
Store the result of this.benchmarks.aggregate.get("Passthrough Copy File") in a variable

fqueze:
This is to ensure the same object is used for .before() and .after() calls, which makes this code work better with https://github.com/fqueze/11ty-fx-profiler/ that replaces the eleventyConfig.benchmarkManager implementation with another one that produces profiles the Firefox Profiler can understand.


zachleat:
Will ship with `3.0.0-alpha.11`, thank you!

PR #3291, opened 2024-05-22
Restore types in package.json and index.d.ts

chriskirknielsen:
Restores https://github.com/11ty/eleventy/pull/2091


zachleat:
Shipping with 3.0.0-alpha.11! Thank you!

PR #3346, opened 2024-07-04
fix: encodeURI for passthrough aliases

satgo1546:
This fixes files with non-ASCII names not working (404) when server passthrough copy behavior is set to `"passthrough"` and the requested file is not already copied to the dist folder.


zachleat (approved):
LGTM, thank you!

zachleat:
Shipping with 3.0.0-alpha.19 and 3.0.0-beta.2

PR #3384, opened 2024-07-27
Refactor to fix TypeScript errors in UserConfig.js

zachleat:
From #3097, related: #3383


zachleat:
Notably this PR removes jsdoc (again) per incompatibilities here: https://github.com/jsdoc/jsdoc/issues/1645

mayank99:
For future reference, this is the correct documentation to look at when using JSDoc comments with `tsc`: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

(https://jsdoc.app/ is different and will cause issues)

zachleat:
Appreciate that! The syntax seems to have diverged enough that I don’t think it’s worth trying to navigate both and I did consciously choose to prefer the TypeScript version moving forward (and removed our jsdoc npm scripts as part of this PR).

Issue #188, opened 2018-08-03
Ignore Files? Ability to do "Draft" builds?


khrome83:
Trying to identify if in `.eleventy.js` if there is a way to determine if I want to process a file or not.

I would love it if I can do the following for example - 

# Frontmatter

```markdown
---
status: draft
---

# Test
```

# .eleventy.js

```javascript
module.exports = function(eleventyConfig) {
  eleventyConfig.ignoreFiles(function(file) {
    return (process.env.DRAFT === 'true') ? false : file.data.status === 'draft';
  });
};
```

# NPM Scripts
```json
"scripts": {
  "build": "DRAFT='true' eleventy --input=src --output=dist",
  "dev": "DRAFT='true' eleventy --input=src --output=dist --serve",
  "publish": "eleventy --input=src --output=dist"
}
```

This would allow the ability to not only setup draft or other conditions for publications, but even future date publications (check timestamp), ignore files to specific output targets - Example - api docs & frontend component docs could share a repo, but have two different build targets? Maybe they end up on different domains. 

I tried looking through the code to see if I could add support for this, but I am a little unclear what determine what should render outside of the ignoreFiles which takes a full file path. So you would need to scan all files in a prestep to add it too that list running the function passed into the config.

zachleat:
Oh, I do like this. The ability to run a filter callback on the found files array before processing would have a lot of utility, I think.

zachleat:
Can you upvote your original post? This is going into the new feature queue

zachleat:
This repository is now using [lodash style issue management](https://twitter.com/samselikoff/status/991395669016436736) for enhancements. This means enhancement issues will now be closed instead of leaving them open.

The enhancement backlog can be found here: https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+

Don’t forget to upvote the top comment with 👍!

khrome83:
@zachleat - love that you adopted the lodash format. First time I am seeing the explanation. Makes perfect sense.

Sort of feel like this should be a GitHub plugin that handles this. Can't wait to see someone monetize the concept. 

Really happy you like the concept. Is there anything short term that would make sense to do besides having a custom collection to build links that are not in draft, and crawling the output and remove pages that were listed in draft? Seems like more work than adding the initial feature, but I am struggling to wrap my head around how the code is organized, no offense.

edwardhorsford:
[...]

PR #3418, opened 2024-08-18
Fix wildcard in extensions argument in `.addPreprocessor()`

vrugtehagel:
The second argument in `.addPreprocessor()` seems to be designed to allow `"*"` to match all templating languages, but gets mapped to `".*"` before the actual comparison to `"*"` happens. As such, it is impossible to match all file types.

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -655,7 +655,7 @@ class Template extends TemplateContent {
 			}
 
 			filters = filters.map((extension) => {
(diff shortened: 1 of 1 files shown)


zachleat (approved):
LGTM

zachleat:
This will ship with 3.0.0-alpha.19 and 3.0.0-beta.2.

zachleat:
Originally at #188

Issue #3415, opened 2024-08-15
The `eleventy.after` event should not fire its handlers in parallel


vrugtehagel:
### Operating system

macOS Sonoma 14.6.1

### Eleventy

3.0.0-beta.1

### Describe the bug

Eleventy should not fire its `eleventy.after` (and probably also not its `eleventy.before`) event handlers in parallel. This is an issue, because any plugin that manipulates files asynchronously now directly conflicts with another, introducting race conditions and potential flaky error scenarios.

For example, I have written a plugin that adds hash parameters to referenced asset URLs. If I were to use it in combination with any plugin that moves or renames files, the ouput would be a mangled mess of half-combinations of files that have moved, some with hash parameters, some without, and potentially duplicate files because one plugin wrote to a file that the other has just moved.

In a nutshell, the [`AsyncEventEmitter`](https://github.com/11ty/eleventy/blob/main/src/Util/AsyncEventEmitter.js#L25) should use a simple for-loop instead of `Promise.all()` to run the handlers sequentially, though potentially we'd want an (internal) option for it to determine whether to run handlers in parallel or not for each individual event type.

zachleat:
Hmm! I think I’m okay with this but it’s too late for this default to ship with 3.0. We could do a temporary feature flag to ease into it.

vrugtehagel:
Cool, I'll draft up a PR for it!

PR #3423, opened 2024-08-25
Add flag to run event handlers sequentially

vrugtehagel:
Resolves https://github.com/11ty/eleventy/issues/3415.

This PR adds a flag to run event handlers sequentially, instead of in parallel. This is to avoid issues with plugins interfering with one another if they are doing asynchronous operations on the same set of files in parallel.

### DIFF
--- a/cmd.cjs
+++ b/cmd.cjs
@@ -31,17 +31,7 @@ const debug = require("debug")("Eleventy:cmd");
 
 	try {
 		const argv = minimist(process.argv.slice(2), {
-			string: [
-				"input",
-				"output",
-				"formats",
-				"config",
-				"pathprefix",
-				"port",
-				"to",
-				"incremental",
-				"event-emitter-mode",
-			],
+			string: ["input", "output", "formats", "config", "pathprefix", "port", "to", "incremental"],
 			boolean: [
 				"quiet",
 				"version",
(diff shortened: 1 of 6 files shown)


zachleat (commented):
Looking great! I definitely don’t want to add a temporary CLI argument for this though—I think it’d be better to keep it as a Configuration API only.

zachleat (changes requested):
I think just revert the changes in cmd.cjs and src/Eleventy.js and we’ll be good

vrugtehagel:
Alright, @zachleat - removed the CLI flag entirely and moved the `setEmitterMode` method to `UserConfig`. So essentially, now users would have the ability to do

```js
export default function(eleventyConfig){
    eleventyConfig.setEmitterMode('sequential');
    // or…
    eleventyConfig.events.setHandlerMode('sequential');
} 
```

Hypothetically, we can remove the former in favor of the latter, but perhaps it's nice to have this "top-level" method since we have it for almost everything else. Let me know what you think :)

zachleat:
@vrugtehagel yeah I think in docs we’ll lean towards only recommending the former method—thank you!

zachleat (approved):
LGTM thank you!

zachleat:
Shipping with 3.0.0-beta.2 and 3.0.0-alpha.19

zachleat:
Temporary docs preview deploying here: https://11ty-website-git-v3-11ty.vercel.app/docs/events/#emitter-modes

Issue #2679, opened 2022-12-06
Change `liquidjs` parameter parsing to use Liquid’s Tokenizer


zachleat:
https://liquidjs.com/tutorials/parse-parameters.html#Parse-Parameters-as-Values
https://liquidjs.com/tutorials/parse-parameters.html#Parse-Key-Value-Pairs-as-Named-Parameters

Currently implementation predated those features and is using a one-off parser using `moo`: https://github.com/11ty/eleventy/blob/0f44f756a30247515033945e74e171a293bc75d5/src/Engines/Liquid.js

I think due to the size of this change, we’ll probably want a configuration escape hatch for folks to swap back to the old parameter parser

Excellent prior art #1263 #1733

AleksandrHovhannisyan:
Hey, just checking to see if you're accepting help/contributions for this issue. I noticed that 11ty's Liquid engine is currently only using moo in two places:

https://github.com/11ty/eleventy/blob/0f44f756a30247515033945e74e171a293bc75d5/src/Engines/Liquid.js#L24

https://github.com/11ty/eleventy/blob/0f44f756a30247515033945e74e171a293bc75d5/src/Engines/Liquid.js#L100

By size of the change, do you mean that it's high impact, or are there other changes that would need to be made as well?

jgerigmeyer:
@zachleat Any thoughts on @AleksandrHovhannisyan's question above? Would you be interested in contributions for this issue? Thanks!

zachleat:
Starting in Eleventy v3.0.0-alpha.18 and newer you can use `eleventyConfig.setLiquidParameterParsing("builtin");` to use the parameter parser provided by Liquid.js. Notably this is an opt-in feature and is not enabled by default in 3.0.

The biggest difference y’all will see with this swap is that with Liquid’s provided tokenizer commas are not allowed to separate parameters. In the one-off parameter parser Eleventy supplied, commas were optional. This commas-optional feature is advertised prominently in the docs: https://www.11ty.dev/docs/shortcodes/

zachleat:
Related: https://github.com/harttle/liquidjs/discussions/724

zachleat:
Temporary docs preview URL deploying here: https://11ty-website-git-v3-11ty.vercel.app/docs/languages/liquid/#shortcode-parameter-parsing

marioortizmanero:
Could the documentation include an example of a Liquid shortcode that uses named parameters? I can't get it to work with:

```
[...]

PR #3444, opened 2024-09-15
Add keyword arguments for Liquid shortcodes

nex3:
This is an adaptation of #1733 to use the built-in Liquid argument parser that was added in #2679. Keyword arguments are passed as an object after all the positional arguments.


zachleat:
Hmm, I’m seeing some test suite failures here from some missing variables in Liquid.js

src/Engines/Liquid.js
```diff
@@ -172,20 +172,57 @@ class Liquid extends TemplateEngine {
...
 		tokenizer.end();
 
 		return parsedArgs;
 	}
 
+	static *evalArguments(tag, ctx) {
+		let argArray = [];
+		let namedArgs = {};
+		if (tag.legacyArgs) {
+			let rawArgs = Liquid.parseArguments(_t.argLexer, tag.legacyArgs);
```
zachleat, line 200: _t is missing here

src/Engines/Liquid.js
```diff
@@ -172,20 +172,57 @@ class Liquid extends TemplateEngine {
...
+			for (let arg of rawArgs) {
+				let b = yield liquidEngine.evalValue(arg, ctx);
```
zachleat, line 202: liquidEngine is missing here

src/Engines/Liquid.js
```diff
@@ -172,20 +172,57 @@ class Liquid extends TemplateEngine {
 		let tokenizer = new Tokenizer(args);
 		let parsedArgs = [];
 
-		let value = tokenizer.readValue();
+		function readValue() {
+			let value = tokenizer.readHash() ?? tokenizer.readValue();
+			// readHash() treats unmarked identifiers as hash keys with undefined
+			// values, but we want to parse them as positional arguments instead.
+			return value?.kind === 64 && value.value === undefined ? value.name : value;
```
zachleat, line 179: I’m seeing some errors here, from LiquidJS’s types: `value` and `name` do not exist on type `HashToken | ValueToken | RangeToken`

zachleat (commented):
This is looking great but has a few small issues!

I think the big question I have is in the test suite tests—I would imagine these tests would need to be gated behind ` eleventyConfig.setLiquidParameterParsing("builtin");` similar to what shipped in https://github.com/11ty/eleventy/issues/2679#issuecomment-2241369136

nex3 (commented):
The tests keep timing out locally, so I'm not 100% confident they're passing now. But I did fix all the known issues!

nex3:
I'm clearly missing something about how to correctly set the parameter parsing...

Issue #3442, opened 2024-09-13
Upgrade to chokidar to 4.0


zachleat:
https://github.com/paulmillr/chokidar/releases/tag/4.0.0

The big difference here is that it drops globs support, which we used both internally and via a public configuration API here: https://www.11ty.dev/docs/watch-serve/#add-your-own-watch-targets

Related #3149

talentlessguy:
11ty depends on `fast-glob`, which can be put on top of chokidar

zachleat:
Related https://github.com/11ty/eleventy/issues/3149

zachleat:
https://github.com/e18e/ecosystem-issues/issues/138

zachleat:
Related https://github.com/11ty/eleventy-dev-server/pull/94 and https://github.com/11ty/eleventy-dev-server/issues/119

zachleat:
This will ship with v4.0.0-alpha.1 (which uses Dev Server v3.0.0-alpha.1 to coordinate the same upgrade across both)

https://github.com/11ty/eleventy-dev-server/releases/tag/v3.0.0-alpha.1

zachleat:
I was able to write a globstripper so that this shouldn’t be a breaking change for our existing APIs https://github.com/11ty/eleventy/blob/deb6dbe62f95545aba80ff4d7bafe5e3d0b72f08/src/Util/GlobStripper.js

PR #3445, opened 2024-09-15
bump chokidar to v4

talentlessguy:
Removes 11 dependencies, includes better types

Passes all tests

Draft for now because needs bumping in dev-server as well

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
 				"@sindresorhus/slugify": "^2.2.1",
 				"bcp-47-normalize": "^2.3.0",
 				"chardet": "^2.0.0",
-				"chokidar": "^3.6.0",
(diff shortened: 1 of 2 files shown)


zachleat:
Sorry—we can’t do this yet per notes made at #3442. Clearly we need some additional test coverage there

talentlessguy:
is it okay if I add tests with a glob matcher (using `fast-glob` which is already a dependency) or it's still a no?

zachleat:
oh, yeah for sure! The plan is to upgrade to v4, internally we rely on that globs feature so we’d need to make sure that doesn’t break when we upgrade. Some more notes on that here: https://www.11ty.dev/docs/watch-serve/#add-your-own-watch-targets

talentlessguy:
how would a test for it look like? should I use fixtures for it?

PR #3509, opened 2024-10-25
Throw error when virtual template format is invalid

dave-kennedy:
It took a while to realize my Atom feed wasn't being created because of this :sweat_smile:


zachleat:
This change will ship with v4.0.0-alpha.1, thank you!

Issue #3518, opened 2024-10-29
Global CommonJS data files in Node 23 do not load their data


vrugtehagel:
### Eleventy

3.0.0

### Describe the bug

When using CommonJS syntax in Node 23, global data files are not loaded properly and results in the data being unavailable throughout the build.

This is because `Module` objects for modules using `module.exports` now contain _two_ keys, `'default'` and `'module.exports'`, and Eleventy specifically [checks if there is only one](https://github.com/11ty/eleventy/blob/main/src/Util/Require.js#L129) before accessing the `default` property. Then it [goes on to make sure](https://github.com/11ty/eleventy/blob/main/src/Util/Require.js#L151-L165) the keys from `'default'` and those in the top level are the same, which fails on the `'module.exports'` key.

zachleat:
Confirmed this one as a bug.

<img width="490" alt="image" src="https://github.com/user-attachments/assets/60e8982b-31dc-41e5-84c1-70ed04925aa7">

My issue here is that Node 23 is an odd version of Node—is there any indication that this change is going to ship for an even version?

vrugtehagel:
Good point @zachleat! I actually don't know if it's here to stay, but it does _kind of_ look like it from [Node's ESM docs](https://nodejs.org/docs/latest/api/esm.html#commonjs-namespaces) (see also the [commit](https://github.com/nodejs/node/commit/1d5ed725e9b0b1d75cd5ff1559f9534287757e71)):

> _These CommonJS namespace objects also provide the `default` export as a `'module.exports'` named export, in order to unambiguously indicate that their representation in CommonJS uses this value, and not the namespace value. This mirrors the semantics of the handling of the `'module.exports'` export name in `require(esm)` interop support._

This not so much specifies whether it is a temporary change or not, but it does seem like Node's attempting to document it properly and has reasons for its existance (which, I'll be honest, I don't quite follow, but they're there).

Either way, given [the fix](https://github.com/11ty/eleventy/pull/3519/files) for it seems rather unintrusive, I think it'd just be a nice-to-have for those on 23 (and potentially beyond) without bothering anyone on Node <23 (I _highly_ doubt anyone's explicitly exporting a `'module.exports'` key).

If you really are worried about this breaking things for users on lower Node versions, we can introduce an extra check against the Node version to make it a Node 23-only change.

zachleat:
I agree, thanks for talking it out! We’ll ship this with v3.0.1-alpha.1

PR #3519, opened 2024-10-29
Fix import-related error with CommonJS modules in Node 23

vrugtehagel:
Fixes https://github.com/11ty/eleventy/issues/3518. Modules in Node 23 (possible 22, didn't test that) have both a `'module.exports'` key as well as a `'default'` key, and Eleventy doesn't know about the former. As such, it incorrectly identifies the module object as the exported object, which breaks things down the road. To resolve this, accept that the top-level object has a `'module.exports'` key that the object under the `'default'` key does not have.


zachleat (approved):

zachleat:
Shipping with v3.0.1-alpha.1

Issue #3460, opened 2024-09-27
Declare subpath exports for bundled plugins, Eleventy-maintained data, etc.


noelforte:
### Is your feature request related to a problem? Please describe.

Upfront note: I'd be willing to submit a PR to help out with this.

Commit c505822 inadvertantly introduced some breaking changes to [subpath exports](https://nodejs.org/api/packages.html#subpath-exports) when Eleventy is consumed as a dependency. By adding the `exports` key to package.json with only value, any attempt to import a subpath throws an error.

### Describe the solution you'd like

This is more specific to `@11ty/eleventy/src/Engines/Util/ContextAugmenter.js` (https://github.com/11ty/eleventy/issues/3355#issuecomment-2316378398) but is applicable to any `exports` in Eleventy that **could** be consumed by outside packages, including (but not limited to):

- Plugins bundled with Eleventy
- `.d.ts` declaration files ([#3097](https://github.com/11ty/eleventy/issues/3097), [#3296](https://github.com/11ty/eleventy/issues/3296))
- Eleventy-maintained global data and methods like `augmentKeys`

Syntax I'm thinking of could be something like:

```json
"exports": {
  ".": {
    "import": "./src/Eleventy.js",
    "require": "./src/EleventyCommonJs.cjs",
  },
  "./user-config": {
    "types": "./types/UserConfig.d.ts"
  },
  "./plugins/html-base": "./src/Plugins/HtmlBasePlugin.js",
  "./plugins/render": "./src/RenderPlugin.js",
  "./plugins/input-path-to-url": "./src/InputPathToUrl.js",
  "./utils/context-augmentor": "./src/Engines/Util/ContextAugmenter.js"
}
```

The above declarations can be consumed as so:

```ts
import type UserConfig from '@11ty/eleventy/user-config';
import Eleventy from '@11ty/eleventy';
import HtmlBasePlugin from '@11ty/eleventy/plugins/html-base';
// ...
```

Or via an aggregator:

```json
"exports": {
  "./internal-plugins": "./src/Plugins/index.js",
}
```

```js
import {
  HtmlBasePlugin,
  RenderPlugin,
  InputPathToUrlPlugin,
} from '@11ty/eleventy/internal-plugins';
```

### Additional context

[...]

PR #3562, opened 2024-11-26
Replace UserConfig typedef with a subpath export (#3460)

noelforte:
This is a minimum viable patch that starts work on #3460. Notably, this change removes the requirement of `maxNodeModuleJsDepth` in `tsconfig.json` for TS packages that want to import UserConfig API, as mentioned [in the mega TS issue](https://github.com/11ty/eleventy/issues/3097#issuecomment-2423050317).

Since modifying the functionality of exports, imports, and declarations will involve some cleanup, removal and renaming, further work is probably something best left to the **next major** release.

I've removed the `@typedef` comment from `Eleventy.js`, since adding a `types` subpath export makes exporting it from Eleventy's main file redundant. If backwards compatiblity is needed, then something like this will implement the fuctionality offered by the subpath export:

```diff
 /**
  * Eleventy’s programmatic API
  * @module 11ty/eleventy/Eleventy
  *
+ * This line is required for IDE autocomplete in config files
+ * @typedef {import('./UserConfig').default} UserConfig
  */
```

Also note, declaring `./src/UserConfig.js` as a `"types"` export will _still block it from being directly consumed_ (see #3460), since `"."` is the only pathname that has a `"import/require"` key. It's therefore only possible to import it as a type, either from Typescript or from JSDoc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,6 @@
 				"minimist": "^1.2.8",
 				"moo": "^0.5.2",
 				"node-retrieve-globals": "^6.0.0",
-				"normalize-path": "^3.0.0",
 				"nunjucks": "^3.2.4",
 				"please-upgrade-node": "^3.2.0",
 				"posthtml": "^0.16.6",
--- a/package.json
+++ b/package.json
@@ -9,8 +9,13 @@
 	"type": "module",
 	"main": "./src/Eleventy.js",
 	"exports": {
-		"import": "./src/Eleventy.js",
-		"require": "./src/EleventyCommonJs.cjs"
+		".": {
+			"import": "./src/Eleventy.js",
+			"require": "./src/EleventyCommonJs.cjs"
+		},
+		"./UserConfig": {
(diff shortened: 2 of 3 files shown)


Ryuno-Ki:
LGTM (as I introduced the original JSDoc comments).

package-lock.json
```diff
@@ -40,7 +40,6 @@
 				"minimist": "^1.2.8",
 				"moo": "^0.5.2",
 				"node-retrieve-globals": "^6.0.0",
-				"normalize-path": "^3.0.0",
```
zachleat, line 43: Does this change belong here? Though I’m on board with it—I don’t think it matters as it is used downstream by `@11ty/eleventy-utils` and will be installed anyway.

zachleat: I’m going to go ahead and push this dep removal independently—ignore my comment above!

zachleat (approved):

zachleat:
Thank you!

PR #3572, opened 2024-12-06
Error stack improvements

noelforte:
Caught an issue while working to improve errors in `eleventy-plugin-vento`. Looks like `EleventyErrorHandler`'s `log` function loop was assigning to `nextRef` instead of the `originalError` key, leading to errors like this:

```njk
{% shortcodeThatThrowsAnError %} <!-- => new Error(...) -->
```

```console
[11ty] Problem writing Eleventy templates:
[11ty] 1. Having trouble rendering njk template ./src/nunjucks.njk (via TemplateContentRenderError)
[11ty] 2. throws an error
```

instead of

```console
[11ty] 1. Having trouble rendering njk template ./src/nunjucks.njk (via TemplateContentRenderError)
[11ty] 2. (./src/nunjucks.njk)
[11ty]   EleventyNunjucksError: Error with Nunjucks shortcode `errorThrow` (via Template render error)
[11ty] 3. throws an error
```

As a related fix, I also added support for the error handler to unwrap `cause` from the Error object if it's defined, even if `cause.originalError` doesn't exist. This will be helpful for custom languages like Vento, which attach more errors to the stack via the `Error.cause` convention.

Would be great to have this in 3.0.1!

(related: https://github.com/noelforte/eleventy-plugin-vento/issues/69)


src/Errors/EleventyErrorHandler.js
```diff
@@ -91,9 +90,9 @@ class EleventyErrorHandler {
 		while (ref) {
 			let nextRef = ref.originalError;
 
-			// Nunjucks wraps errors and puts the original in error.cause
-			if (nextRef?.cause?.originalError) {
-				nextRef = nextRef.cause.originalError;
+			// Unwrap cause from error and assign it to what Eleventy expects
```
Ryuno-Ki, line 93: Have you looked at

https://github.com/11ty/eleventy/commit/9f957abb624270bc33fcc1579436c7f771c83c86#diff-bf3ab5639a6c77684327c6f227d9ea440e361ebfb365ebfe93effde8c80c4926

to make sure it behaves well with Nunjucks?

Or, what template engines have you tested this with?

noelforte: > Have you looked at https://github.com/11ty/eleventy/commit/9f957abb624270bc33fcc1579436c7f771c83c86#diff-bf3ab5639a6c77684327c6f227d9ea440e361ebfb365ebfe93effde8c80c4926 to make sure it behaves well with Nunjucks?

See my original comment—the example I'm using **is** Nunjucks. This change improves the error; previously the error specifying `Error with Nunjucks shortcode...` wasn't getting logged.

I also just tested with JavaScript and Liquid templates—this change doesn't degrade logging of their errors at all (although the line referencing this is a "shortcode error" doesn't get logged for them either). This edit only improves Nunjucks and Custom templates that chain errors via `cause`.

FFIW, as of Node 18 (Eleventy's minimum supported version), [the convention is to rethrow errors with `cause` passed as part of the `ErrorOptions` object](https://nodejs.org/docs/latest-v18.x/api/errors.html#new-errormessage-options). That said, should Eleventy's error handler be refactored to align with that standard instead of the non-standard `.orignalError` property currently used?

Ryuno-Ki: > That said, should Eleventy's error handler be refactored to align with that standard instead of the non-standard .orignalError property currently used?

That's for @zachleat to decide. (Perhaps worth a new issue?)

zachleat: Absolutely, filed that here: #3582

src/Errors/EleventyErrorHandler.js
```diff
@@ -77,7 +76,7 @@ class EleventyErrorHandler {
 	//https://nodejs.org/api/process.html
 	log(e, type = "log", chalkColor = "", forceToConsole = false) {
 		if (process.env.DEBUG) {
-			debug("Full error object: %o", util.inspect(e, { showHidden: false, depth: null }));
+			debug("Full error object: %O", e);
```
zachleat, line 79: This change is not quite the same, right? https://github.com/debug-js/debug/blob/7e3814cc603bf64fdd69e714e0cf5611ec31f43b/src/node.js#L262 looks like it uses util.inspect defaults, which is `depth: 2`

https://nodejs.org/docs/latest/api/util.html#utilinspectobject-showhidden-depth-colors

zachleat: Gonna roll forward without this line!

zachleat (dismissed):

zachleat (approved):

zachleat:
This will ship with v3.0.1-alpha.1 thank you!

PR #3614, opened 2025-01-08
await mdlib.render

boehs:
Hi Zach, I'm over here again with my incredibly-overengineered-11ty-fuzzing-how-tf-does-this-monstrosity-work-website-called-boehs.org, and this is a tiny patch.

Markdown-it does not support async, but you can implement your own markdown renderer by `{ render: (fileContents,data) => string }`, and you may wish for this renderer to be asynchronous because you commit [coding horrors](https://github.com/boehs/site/pull/42/files)

Javascript spec says

`[rv] = await expression;`
expression: A Promise or any value to wait for.
rv: Returns the fulfilled value of the promise, *or the value itself if it's not a Promise.*

so you can kinda just chuck an await on there and pray for the best.

### DIFF
--- a/src/Engines/Markdown.js
+++ b/src/Engines/Markdown.js
@@ -72,7 +72,7 @@ class Markdown extends TemplateEngine {
 				return async function (data) {
 					let fn = await fnReady;
 					let preTemplateEngineRender = await fn(data);
-					let finishedRender = mdlib.render(preTemplateEngineRender, data);
+					let finishedRender = await mdlib.render(preTemplateEngineRender, data);
 					return finishedRender;
 				};
 			}


zachleat:
HMM! If the function there is already async isn't it returning a promise or no? Shouldn't this already work?

Ryuno-Ki:
You can directly return the `mdlib.render` call here.

Yes, it is always returning a Promise.
Using an additional async here should [add another entry to the microtask queue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#task_queues_vs._microtasks).

(I was expecting to find a link on [What Color is Your Function](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) here :sweat_smile: )

boehs:
> HMM! If the function there is already async isn't it returning a promise or no? Shouldn't this already work?

i'm so silly

zachleat:
@boehs don't you dare! *Everyone* makes mistakes (especially me) and that's what code review is for! Keep those PRs coming!

Issue #3631, opened 2025-01-23
Dev server broken on Windows


kate-bagenzo:
### Operating system

Windows 10

### Eleventy

3.0.0

### Describe the bug

The dev server is inconsistent and not logging most changes

. Sometimes it works for one or two edits, before requiring manual stopping and restarting. Deleting a file and re-adding it causes the server to crash, with what appears to be a Windows path normalization error:
```
Output conflict: multiple input files are writing to ./_site/posts/post1/index.html. Use distinct permalink values to resolve this conflict.
[11ty]   1. ./src\posts\post1.md
[11ty]   2. ./src/posts/post1.md (via DuplicatePermalinkOutputError)
```
Have encountered this while using [11ty base blog](https://github.com/11ty/eleventy-base-blog) but can also reproduce with a blank site as described below. Clearing `_site` does not seem to have an effect. Does not appear on Linux.

[log.txt](https://github.com/user-attachments/files/18525858/log.txt)

### Reproduction steps

1. Initialize a blank 11ty project with `npm init -y` and `npm i @11ty/eleventy`
2. Create an `index.md` and a `src/posts/test.md`
3. Start the dev server with `npx @11ty/eleventy --serve`
4. Delete the file `test.md`
5. Create the file again
6. Dev server stops with attached error

### Expected behavior

Dev server responds to changes, does not crash on reading deleted files

### Screenshots

![Image](https://github.com/user-attachments/assets/fa84f2eb-8750-408c-b021-51fb3e91f141)

zachleat:
Successfully reproduced this one on Windows!

PR #3632, opened 2025-01-23
fix: inconsistent filepath separators

seleb:
Hey there! Ran into the issue described in #3631 where filepaths are being formatted inconsistently on a windows dev machine, and this seems to prevent the dev server from updating to show new content.

Sample result from `_globSearch` when the server is started with `test1.md`, and then a new file `test2.md` is added:
```js
[
  './content/blog/test1.md',
  './content\\blog\\test2.md',
]
```

The inconsistent format in the newly added file seems to be carried through all the way into the templates without ever being normalized.

This is a bit of a naive fix, but not sure if there's a better place upstream to place it. I'm not familiar with the specifics but it looks like [historically](https://github.com/11ty/eleventy/commit/7e477f27b833da49ed37ec05733be457ac05c595#diff-8b8317c75f11cacab005f253a162744e6cf3530292234efec197702d27540680L204) there was transformation between `templateGlobs` and `normalizedTemplateGlobs` which may be related?

It's also possible this is a bug in `fast-glob`, which appears to be responsible for performing the actual search, but as this causes issues with the 11ty dev server I figured it'd be worth raising a PR where there's a known user-facing issue/possible fix and let the maintainers make that call.

### DIFF
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -334,9 +334,10 @@ class EleventyFiles {
 
 		// returns a promise
 		debug("Searching for: %o", globs);
-		return (await this.fileSystemSearch.search("templates", globs, {
+		const results = await this.fileSystemSearch.search("templates", globs, {
 			ignore: this.uniqueIgnores,
-		})).map(i => i.replace(/\\/g, '/'));
+		});
+		return results.map(i => i.replace(/\\/g, '/'));
 	}
 
 	getPathsWithVirtualTemplates(paths) {
--- a/src/EleventyFiles.js
+++ b/src/EleventyFiles.js
@@ -329,14 +329,14 @@ class EleventyFiles {
 		return ret;
 	}
 
-	_globSearch() {
+	async _globSearch() {
 		let globs = this.getFileGlobs();
 
 		// returns a promise
 		debug("Searching for: %o", globs);
-		return this.fileSystemSearch.search("templates", globs, {
+		return (await this.fileSystemSearch.search("templates", globs, {
 			ignore: this.uniqueIgnores,
-		});
+		})).map(i => i.replace(/\\/g, '/'));
 	}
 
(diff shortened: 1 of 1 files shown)


src/EleventyFiles.js
```diff
@@ -329,14 +329,14 @@ class EleventyFiles {
...
+	async _globSearch() {
 		let globs = this.getFileGlobs();
 
 		// returns a promise
 		debug("Searching for: %o", globs);
-		return this.fileSystemSearch.search("templates", globs, {
+		return (await this.fileSystemSearch.search("templates", globs, {
 			ignore: this.uniqueIgnores,
-		});
+		})).map(i => i.replace(/\\/g, '/'));
```
Ryuno-Ki, line 339: Using [`path.sep`](https://nodejs.org/api/path.html#pathsep) would work here, too?

Instead of an IIFE for the promise I think it's okay to add another temporary variable.

```suggestion
		const results = await this.fileSystemSearch.search("templates", globs, {
			ignore: this.uniqueIgnores,
		});
		return results.map(i => i.replace(new RegExp(path.sep, 'g'), '/'));
```

Untested, as I'm on Debian Linux.

seleb: I'll add the temporary variable for clarity, but `path.sep` would not work here without being escaped. I don't think there'd be any benefit in using it either: replacing linux path separators would not do anything.

zachleat:
Testing a `issue-3632` branch that makes use of our `TemplatePath.standardizeFilePath` utility from `@11ty/eleventy-utils` for this purpose:

Test run https://github.com/11ty/eleventy/actions/runs/13642618315
Issues https://github.com/11ty/eleventy/compare/issue-3632

zachleat:
After a deeper dive into #3631 this wasn’t the root cause, though I appreciate the help and will still apply the (modified) change you suggested!

Issue #3582, opened 2024-12-10
Housekeeping: use error.cause everywhere in Eleventy


zachleat:
Maybe get rid of EleventyBaseError and all the `originalError` stuff.

Supported by Node 16.19.0+

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

via #3572 cc @noelforte and @Ryuno-Ki

zachleat:
Included with #3858

Issue #2328, opened 2022-04-15
Custom template languages and "Tried to use templateContent too early" errors.


johnridesabike:
I'm working on a custom template language, and my site kept failing to compile due to "Tried to use templateContent too early" errors. This was puzzling, because a very similar version of the site would compile just fine with a different template language, and I couldn't figure out why `templateContent` was possibly being used too "early."

I finally figured it out, I think. Here is my understanding of what was happening:

- In my custom language extension's `compile` function, it catches exceptions, logs their messages, and then throws another exception.
- When `templateContent` is first accessed, it isn't rendered yet and throws this error. Eleventy will catch this error and then try to render the `templateContent` field.
- Because I was catching the "templateContent too early" exception, and throwing my own, this broke Eleventy.

Once I started re-throwing the *same* exception that I caught, then the site suddenly started working again. (This is fine, since I had no real reason to throw a different exception, anyway.)

There is currently no documentation about this behavior, and the exception's message doesn't communicate what's going wrong. My suggestion is: can we add something to the "Custom template extension" documentation page, warning people not to catch exceptions that come from Eleventy, perhaps even mentioning this situation as a "common pitfall" (if it's common)?

Perhaps there's a better way to implement this `templateContent` feature that doesn't leak unexpected behavior, but I think we should at least document the existing details.

zachleat:
Good call-out!

Related: #1615 and #975

zachleat:
#975 is shipping with 2.0.0-canary.20

PR #3651, opened 2025-02-05
Check `Error.cause` for `TemplateContentPrematureUseError` when rendering

noelforte:
This PR may address the issue of https://github.com/11ty/eleventy/issues/2328, and resolves a downstream issue in https://github.com/noelforte/eleventy-plugin-vento/issues/120.

I've patched the static method `EleventyErrorUtil.isPrematureTemplateContentError` to search for instances of `TemplateContentPrematureUseError`s within `Error.cause` as well, so rethrown errors (from a Custom Template Language in my case) are properly caught by Eleventy.

As work on https://github.com/11ty/eleventy/issues/3582 continues, the implementation here could probably be cleaned up even more but I figured I'd start with this single patch.

I've also added additional changes in separate commits for the following:

1. Refactored the conditional to leverage optional chaining (`?.`)
2. Removed extra checks on `originalError.name` that didn't seem to have any effect since `originalError.originalError` is what we're actually looking to identify. Unless of course `TemplateContentPrematureUseError`s can be thrown in cases that are not `RenderError`s or `UndefinedVariableError`s, in which case you can drop that commit.

Let me know if there's any changes!

### DIFF
--- a/src/Errors/EleventyErrorUtil.js
+++ b/src/Errors/EleventyErrorUtil.js
@@ -60,12 +60,11 @@ class EleventyErrorUtil {
 		// TODO the rest of the template engines
 		return (
 			e instanceof TemplateContentPrematureUseError ||
-			(e.originalError &&
-				(e.originalError.name === "RenderError" ||
-					e.originalError.name === "UndefinedVariableError") &&
-				e.originalError.originalError instanceof TemplateContentPrematureUseError) || // Liquid
-			(e.message || "").indexOf("TemplateContentPrematureUseError") > -1
-		); // Nunjucks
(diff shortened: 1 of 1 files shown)


noelforte:
Hey @zachleat sorry to bug you—is there any way I could get a review on this and potentially merged before 3.0.1 goes out the door?

zachleat:
Happy to merge this one though I did restore the Liquid.js specific name checks (though their removal broke no tests, just playing it safe there)

zachleat (approved):

zachleat:
Shipping with v3.0.1-alpha.5

Issue #3654, opened 2025-02-10
No error is generated for an incorrectly spelled default config filename.


JessyCatterwaul:
### Is your feature request related to a problem? Please describe.

I'm a beginner to most of the technologies involved with Eleventy, and was following this tutorial: https://learn-eleventy.pages.dev/lesson/3/
I received errors when running `npx eleventy --serve`, that I did not understand. 

I was helped by what seems to be an amazing Discord community. 😺
It turns out that I had just misspelled `eleventy.js` as `elventy.js`.

With `eleventy` not being a word that I'm all that familiar with looking at, that didn't register to my brain.

### Describe the solution you'd like

I'm sorry, I don't know what is practical here. I just would have liked to have seen something automated that told me what the problem was, rather than

```
[11ty] 1. Problem creating an Eleventy Layout for the "./src/index.md" template file. (via EleventyBaseError)
[11ty] 2. You’re trying to use a layout that does not exist: _includes/layouts/home.html (via `layout: layouts/home.html`)
```

or 

```
[11ty] 1. Having trouble rendering liquid template ./src/_includes/layouts/home.html (via TemplateContentRenderError)
[11ty] 2. tag "extends" not found, file:./src/_includes/layouts/home.html, line:1, col:1 (via ParseError)
[11ty] 3. tag "extends" not found (via AssertionError)
```

vrugtehagel:
I agree that this is an issue, and I've seen some people running into issues creating a config file with a name that _seems_ like it would work but doesn't (like `.eleventy.mjs` or `.eleventy.cjs`).

My suggestion is to generate a warning message if no configuration file was found at all - I suspect most Eleventy projects have one anyway. It's not ideal for projects that (intentionally) don't have one, but I'm guessing that such a message would save more headaches than it creates regardless.

zachleat:
Just for completeness we _do_ throw an error if a `--config` CLI was specified and not found, which I just confirmed:

```
/Temp/eleventy-3654 ᐅ npx @11ty/eleventy --config=eleventy.js       
[11ty] Eleventy Error (CLI initialization):
[11ty] A configuration file was specified but not found: eleventy.js (via EleventyConfigError)
[11ty] 
[11ty] Original error stack trace: EleventyConfigError: A configuration file was specified but not found: eleventy.js
[...]

PR #3675, opened 2025-03-05
Add warning for missing default Eleventy config

vrugtehagel:
Creates a warning for when a default config file (e.g. `.eleventy.js` or `eleventy.config.js`) was not found and none was explicitly specified using the `--config` option.

Fixes https://github.com/11ty/eleventy/issues/3654.

For the wording of this error, I avoided saying "No default config file found" since following that up with "using the default configuration" would be confusing. It also doesn't seem necessary to say "default config file" anyway, since this warning cannot appear when a config file was manually specified, so there is no ambiguity.

### DIFF
--- a/src/TemplateConfig.js
+++ b/src/TemplateConfig.js
@@ -365,6 +365,8 @@ class TemplateConfig {
 			throw new EleventyConfigError(
 				"A configuration file was specified but not found: " + this.projectConfigPaths.join(", "),
 			);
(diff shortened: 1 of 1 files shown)


zachleat:
Going to close this one per comment left on https://github.com/11ty/eleventy/issues/3654#issuecomment-2810902240 sorry! Happy to discuss other options!

PR #3689, opened 2025-03-12
perf: dynamic import dev server

### LINKED ISSUE #3621: performance yak shaving
Hi, I was doing a little [cpu profiling](https://github.com/boehs/site/pull/46) as one does when they have nothing better to do, 

## dev server

and I noticed that 11ty seems to be burning precious milliseconds on loading the development server even when the serve command is not called.

I noticed that 50-60ms appears to be in `(anonymous server.js:1) (file:///Users/evan/Code/site/node_modules/.pnpm/@11ty+eleventy-dev-server@2.0.4/node_modules/@11ty/eleventy-dev-server/server.js:1:1)`

The command I did was `node --cpu-prof ./node_modules/@11ty/eleventy/cmd.cjs --quiet`, so no dev server should have launched, and indeed none did, but node still spent time loading it due to the top level import. `await import(`ing would save 50-60ms or so. Amazing!!

## ~~micromatch and fast-glob~~

~~it appears 11ty is using two glob libraries for some reason? each one of them takes between 10-15ms to load, so that much would be saved by only using one~~

maybe the micromatch direct dep can be removed but it won't impact perf because fast-glob uses it

thank you for attending my ted talk. god bless 11ty.

boehs:
closes #3621 

cpuprof showed significant time spent loading libraries for the dev server. This ensures it is only imported when needed

```
~/Code/small/eleventyb: hyperfine "pnpm build" --warmup 1
Benchmark 1: pnpm build
  Time (mean ± σ):     636.3 ms ±  88.6 ms    [User: 572.3 ms, System: 92.2 ms]
  Range (min … max):   582.1 ms … 878.8 ms    10 runs
// after change
~/Code/small/eleventyb: hyperfine "pnpm build" --warmup 1
Benchmark 1: pnpm build
  Time (mean ± σ):     593.4 ms ±  49.5 ms    [User: 551.8 ms, System: 88.3 ms]
  Range (min … max):   533.4 ms … 701.4 ms    10 runs
```

image

tested by verifying `eleventy --serve` still works

assuming 50ms of savings, with 60k downloads per week, eleventy is likely called 100k times per week, or `5200000` times per year. This means this change will save 72 hours of compute per year! buying puts on AMD!!!!!

### DIFF
--- a/src/EleventyServe.js
+++ b/src/EleventyServe.js
@@ -2,7 +2,6 @@ import assert from "node:assert";
 
 import debugUtil from "debug";
 import { Merge, DeepCopy, TemplatePath } from "@11ty/eleventy-utils";
-import EleventyDevServer from "@11ty/eleventy-dev-server";
 
 import EleventyBaseError from "./Errors/EleventyBaseError.js";
 import ConsoleLogger from "./Util/ConsoleLogger.js";
@@ -84,7 +83,7 @@ class EleventyServe {
 	async getServerModule(name) {
 		try {
 			if (!name || name === DEFAULT_SERVER_OPTIONS.module) {
-				return EleventyDevServer;
+				return import("@11ty/eleventy-dev-server").then(i=>i.default)
 			}
 
 			// Look for peer dep in local project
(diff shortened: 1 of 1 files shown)


zachleat (approved):
LGTM, thank you very much!

zachleat:
Shipping with v3.0.1-alpha.5

Issue #3481, opened 2024-10-10
Bug Report: Circular Structure Error in Hot Reload with .11ty.js Files


teamchan3:
### Operating system

maxOS 15.0.1

### Eleventy

version: 3.0.0

### Describe the bug

I encountered a build error related to circular structure when using .11ty.js files. The error occurs specifically during hot reloading after editing the file.

Steps to Reproduce:

1. Create a new index.11ty.js file with the following content:
`export default "<p>aaa</p>";`

2. Run Eleventy in serve mode:
`npx @11ty/eleventy --serve`

3. Edit and save index.11ty.js in an editor.
4. The following error is thrown in the console, and the hot reload fails:
`reload-client.js:21 [11ty][02:36:03.136 UTC] Build error: Converting circular structure to JSON`

Expected Behavior:

The hot reload should work without throwing an error when editing .11ty.js files.

Actual Behavior:

When editing and saving the file, a “Converting circular structure to JSON” error is thrown, and hot reload does not work.

Error Log:

```
reload-client.js:21 [11ty][02:36:03.136 UTC] Build error: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'page' -> object with constructor 'Object'
    --- property 'rawInput' closes the circle
{stack: 'TypeError: Converting circular structure to JSON\n …de_modules/@11ty/eleventy/src/Eleventy.js:1207:4)', message: "Converting circular structure to JSON\n    --> star…ct'\n    --- property 'rawInput' closes the circle"}
```

•	Eleventy version: 3.0.0
•	Node.js version: 22.1.0
•	OS: maxOS 15.0.1

This issue only seems to occur with .11ty.js files; it does not happen with .njk files. It appears related to the hot reload process involving reload-client.js.

I would appreciate any guidance on how to resolve this or whether a fix is required.

This format should provide enough clarity and detail for the Eleventy team to investigate the issue.

zachleat:
Can anyone else reproduce this one? I tried with the instructions provided and was unable to reproduce.

ccoloma:
[This repo](https://github.com/ccoloma/11ty-with-ts) uses TS (I created it to address another issue related to caching dependencies), but you can reproduce the current error here. Just launch `npm run watch` and update `src/pages/index.11ty.tsx`.

abstractvector:
[...]

Issue #3348, opened 2024-07-04
Add support for `rawInput` for `11ty.js` templates.


zachleat:
#1206

zachleat:
> By default this repo closes enhancement request issues and places them in a queue for folks to upvote. We do this to avoid a large build-up of open issues. We got this idea from the [lodash project](https://twitter.com/samselikoff/status/991395669016436736).

> Don’t forget to upvote the first comment with 👍 to register your vote! You can view [the current enhancement backlog](https://github.com/11ty/eleventy/issues?utf8=%E2%9C%93&q=label%3Aneeds-votes+sort%3Areactions-%2B1-desc+).

zachleat:
Note that `rawInput` is available for virtual 11ty.js templates #3347

Issue #1206, opened 2020-05-22
Access to raw file content (before any processing)


sjml:
I know from the `page` object I can access `content` and within a collection I can use `templateContent` to get the rendered output, but is it possible to get just the markdown input? 

I could just read the file from the page's `inputPath`, but then I also get the frontmatter... it's easy enough to make a quick shortcode to return what I'm after, but I wanted to make sure there wasn't something built-in that I was missing...

MiloCasagrande:
I think `frontMatter.content` holds the markdown, without the frontmatter part.

nhoizey:
@MiloCasagrande I didn't see that anywhere, would you have an example?

MiloCasagrande:
@nhoizey If you meant an example in the docs, I didn't see any in there.
This is what gives me the markdwon from within a collection. I assume that whatever the looping mechanism the same data is available:
```
---
pagination:
  data: collections.post
  size: 10
  alias: posts
---
{% for post in posts %}
  {{ post.template.frontMatter.content | log }}
{% endfor %}
```

nhoizey:
Thanks, it works in loops indeed! 👍

But I can't make it work in a content's template. Any idea how to access it?
I tried `page.template.frontMatter.content` and multiple others…

Also, I guess it should be documented, if we want it to keep working in the future… 😅

MiloCasagrande:
@nhoizey unfortunately not, I was only able to access it from within the pagination loop.

Those data structures should be documented somewhere, I also fear that they might disappear in the future.

nhoizey:
Ok.

@zachleat can we expect this to stay in the future? Also, is there a way to access the same value in a content's template/layout?

zachleat:
Another vote for this here https://twitter.com/rob_dodson/status/1321554550671618048 from @robdodson

robdodson:
yep, here's how we use this in web.dev to build our algolia index
https://github.com/GoogleChrome/web.dev/blob/master/src/site/_data/lib/algolia.js#L62

but I feel a little nervous using an undocumented API 😅

aarongustafson:
Interesting, thanks for sharing @robdodson. We are about to use the following in a custom collection for https://github.com/webwewant/webwewant.fyi:

```js
  eleventyConfig.addCollection("wantsBySubmissionDate", collection => {
    return collection.getAll()
             .filter( item => item.inputPath.match(/\/wants\/.*\.md/) !== null )
[...]

PR #3693, opened 2025-03-13
Prevent rawInput circular structure

knokmki612:
JavaScript._getInstance(mod) sometime takes data object, will be circular. This change prevents passing circular object to dev server.

Probably related to https://github.com/11ty/eleventy/issues/3481 https://github.com/11ty/eleventy/issues/1206 https://github.com/11ty/eleventy/issues/3348

---

My code reading notes (Sorry if I misunderstood about implementation. I have no confidence.)

> JavaScript._getInstance(mod) sometime takes data object

https://github.com/11ty/eleventy/blob/2f559c7bd4b722b6927e31f96baad50e73ab90a9/src/Engines/JavaScript.js#L76-L84

> passing circular object to dev server

https://github.com/11ty/eleventy/blob/2f559c7bd4b722b6927e31f96baad50e73ab90a9/src/Eleventy.js#L984-L996

### DIFF
--- a/src/Template.js
+++ b/src/Template.js
@@ -733,7 +733,9 @@ class Template extends TemplateContent {
 		}
 
 		// https://github.com/11ty/eleventy/issues/1206
-		data.page.rawInput = rawInput;
+		data.page.rawInput = JSON.parse(
+			JSON.stringify(rawInput, (_, v) => (typeof v === "function" ? v.toString() : v)),
+		);
 
 		if (!Pagination.hasPagination(data)) {
 			await this.addComputedData(data);


src/Template.js
```diff
@@ -733,7 +733,9 @@ class Template extends TemplateContent {
 		}
 
 		// https://github.com/11ty/eleventy/issues/1206
-		data.page.rawInput = rawInput;
+		data.page.rawInput = JSON.parse(
+			JSON.stringify(rawInput, (_, v) => (typeof v === "function" ? v.toString() : v)),
+		);
```
knokmki612, line 738: Or would it be better to simply not return anything...?

```suggestion
		data.page.rawInput = data.page.templateSyntax != "11ty.js" ? rawInput : undefined;
```

zachleat:
Appreciate the research here! I’ll leave a comment on #3481 with more detail but we want to keep `rawInput` available for use in templates and I don’t believe it’s causing issues outside of the dev server serialization so we’ll likely solve it in the server code (instead of modifying `rawInput`).

Just as an FYI, collection items are already not JSON serializable (https://www.11ty.dev/docs/collections/#collection-item-data-structure) because of circular refs in `data.collections`

PR #3709, opened 2025-03-17
HtmlBasePlugin: Fix confusing error message

noelleleigh:
Fixed issue where an error message referred to the wrong option name (`base` vs. `baseHref`), leading to confusion.


zachleat (approved):
Thank you!

PR #3715, opened 2025-03-20
Dependency optimization

outslept:
Replace several dependencies with more lightweight alternatives and native Node.js functionality:

- Replace `p-map` with custom `Promise.all` implementation using chunking
- Replace `graceful-fs` with native `node:fs` module (Node.js >=18 provides sufficient capabilities)
- ~~Replace `minimist` with `node:util.parseArgs()` for CLI argument parsing~~
- Replace `husky` + `lint-staged` with `simple-git-hooks` + `nano-staged` (fewer sub-dependencies)

These changes reduce the deps size, lower the overall dep count, and hopefully improve maintainability while preserving identical functionality. Part of the: https://github.com/e18e/ecosystem-issues/issues/164

UPD: I'm open to any change/reverts you'd like for this


outslept:
Also here are the videos with coverage result for the fork:

https://github.com/user-attachments/assets/50f0c483-64a8-41b9-8ad0-db6c52880316

And for the current 11ty version

https://github.com/user-attachments/assets/4820cd23-f342-4938-b2d9-fbbcb792a16c

src/Util/FileSystemManager.js
```diff
@@ -1,11 +1,9 @@
 import path from "node:path";
 import fs from "node:fs";
-
-import gracefulFs from "graceful-fs";
 import util from "node:util";
 
-const mkdir = util.promisify(gracefulFs.mkdir);
-const writeFile = util.promisify(gracefulFs.writeFile);
+const mkdir = util.promisify(fs.mkdir);
```
Ryuno-Ki, line 5: You can skip the promisifying here by importing from `node:fs/promises`:

https://nodejs.org/docs/latest/api/fs.html#fspromisesmkdirpath-options

src/Util/FileSystemManager.js
```diff
@@ -1,11 +1,9 @@
...
+const writeFile = util.promisify(fs.writeFile);
```
Ryuno-Ki, line 6: You can skip the promisifying here by importing from `node:fs/promises`:

https://nodejs.org/docs/latest/api/fs.html#fspromiseswritefilefile-data-options

outslept: Done

src/Util/FileSystemManager.js
outslept: Done

zachleat:
Happy to merge this but can you move the minimist->parseArgs change to a separate PR? I don’t think parseArgs supports multi-type CLI args yet (boolean or string values) — and this is the riskiest change of the 3 I think

zachleat:
(I fixed the conflicts I made here too)

outslept:
> Happy to merge this but can you move the minimist->parseArgs change to a separate PR? I don’t think parseArgs supports multi-type CLI args yet (boolean or string values) — and this is the riskiest change of the 4 I think

Sure thing, gimme a minute

outslept:
Oki, I've just copied the code from https://github.com/11ty/eleventy/blob/main/cmd.cjs Some of the tests seem to be failing locally. Does this need investigation

image

Ryuno-Ki:
`path.relative()` sounds like a sensible choice to me.

Maybe we can back out this piece into a separate PR to get the majority in here, though.

zachleat (approved):

zachleat:
LGTM this will be in v3.0.1-alpha.6 (which we will likely end up as stable in v3.1.0 given it’s growing size)

thank you!!

Issue #3652, opened 2025-02-05
Extract some core features to official plugins


uncenter:
In addition to #3647, I think Eleventy should avoid including opinionated and complicated features like content dates and tags to core. Tags create issues like https://github.com/11ty/eleventy/issues/1791 (see also https://github.com/11ty/eleventy/issues/1254, which proposes removing this feature as well), content dates lead to us relying on the large Luxon library (https://github.com/11ty/eleventy/issues/3587), etc. This kind of functionality would be better suited to plugins; `@11ty/eleventy-plugin-tags` and `@11ty/eleventy-plugin-content-dates`.

zachleat:
Cross linking Bluesky discussion here https://bsky.app/profile/uncenter.dev/post/3lj4fvnbpdk26

zachleat:
Probably a year or two away yet but the long term vision will likely be to use the Temporal API instead of luxon, eventually! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

zachleat:
Going to change this one to be exclusive to tags/collections. The dates issue is #3587

Issue #3647, opened 2025-02-03
Extract "default" template languages (Nunjucks, Liquid, markdown) to plugins


uncenter:
> ...like the more esoteric templating languages that were moved out of core to plugins with v3, Nunjucks and Liquid should also be plugins - so that I, someone who uses neither of those two (I use [Vento](https://github.com/noelforte/eleventy-plugin-vento)), am not installing an extra 3MB every time. 

 _Originally posted by @uncenter in [#3587](https://github.com/11ty/eleventy/issues/3587#issuecomment-2629484990)_

---

As suggested by @Ryuno-Ki, opening as a new issue for this discussion. Since Eleventy tries to remain unopinionated, I think moving these template languages to separate plugins in https://github.com/11ty/eleventy-plugin-template-languages would make sense. If it is felt that including (a) default template language(s) is important (i.e. new users, simplified configuration/setup - valid points), I would propose splitting Eleventy into two packages; `@11ty/core` and `@11ty/eleventy`, the former being a more barebones scaffolding tool and the latter coming with more defaults and reinforced structure. Just a thought :)

Ryuno-Ki:
Thank you 🧡 

I think, one of the beautiful aspects of Eleventy is the support for multiple template languages (as well as their mix-and-match). This eases the transition from other systems to it (Jekyll using Liquid, Python with Jinja2 which is almost identical to Nunjucks etc).

On the other hand there's this argument that by bundling all those template language engines the package download size goes up.

Splitting it up might be a good strategy. In the past, several pieces have been moved into plugin land.

boehs:
In regards to forming a core, eleventy has a lot of odd vestigial organs supporting languages like HTML, Markdown, Nunjucks and Liquid inside of core. This includes but is not limited to the following:

- `setLiquidOptions` / `addLiquidFilter` / `setLiquidParameterParsing` / `addLiquidShortcode`
- `setNunjucksEnvironmentOptions` / `addNunjucksFilter` / `addNunjucksAsyncFilter` / `addNunjucksShortcode` / `addPairedNunjucksShortcode` / `addPairedNunjucksAsyncShortcode` / `addNunjucksGlobal`
- `markdownTemplateEngine`

all of these functions internally either

- are a thin wrapper for another API
- *should* be a thin wrapper for another API

[...]

PR #3747, opened 2025-04-08
Begin process of decoupling all templating languages

boehs:
This pull request, codeveloped with @uncenter, dramatically simplifies the API surface of eleventy by deprecating dozens of language specific functions. The full scope is unknown, over time this body will be amended.

The guiding vision of this pull request is to move towards a future where eleventy core does not make assumptions about templating languages, however the default experience is still just as powerful (so users opt out of additional features, not in). Why somebody might wish to do this:

1. Free themselves from magic defaults
2. Improve build times

The first step, which is this pull request, is to get eleventy to a place where nunjucks and other templating languages could conceivably be implemented as a plugin. Even if the broader goal is not achieved, this pull request stands great on its own by enhancing modularity and extendability.

## The PR

In cases where required, the functionality of generic APIs has been extended in backwards compatible manners, and internally, all deprecated functions have been refactored to use generic & public counterparts.

We’ve attempted to ensure that as part of the refactor of deprecated functions, the documented functionality remains the same. We expect that unholy code relying on undocumented features may break.

## Filters

Bad code on top of Nunjuck’s bad async code may break in different and unpredictable ways.

- [x] Deprecates `addNunjucksFilter` in favour of `addFilter(..,..,{langs: ["njk"]})`
- [x] Deprecates `addAsyncNunjucksFilter` in favour of `addFilter(..,..,{langs: ["njk"]})`
- [x] Deprecates `addLiquidFilter` in favour of `addFilter(..,..,{langs: ["liquid"]})`
- [x] Adds a `langs` attribute to `addFilter` so that languages can filter by lang in `getFilters`

Note that one limitation is that you can’t have two different implementation of filters with the same name targeting different languages. The same holds true for shortcodes. I can think of some crafty workarounds but I struggle to think of cases where this would be desired. 

## Shortcodes

- [ ] Deprecates `addNunjucksShortcode`
- [ ] Deprecates `addNunjucksAsyncShortcode`
- [ ] Deprecates `addPairedNunjucksAsyncShortcode`
- [ ] Deprecates `addPairedNunjucksShortcode`
- [ ] Deprecates `addLiquidShortcode`

## JavaScript

- [x] Removes `addJavaScriptShortcode`
- [x] Removes `addPairedJavaScriptShortcode`
- [x] Removes `addJavaScriptFilter`
- [ ] Deprecates `addJavaScriptFunction` in favour of `addFilter` (not done yet; tbd)

With JavaScript, we're outright removing a number of functions that do exactly the same thing and aren't publicly documented.

## Misc

- [ ] Deprecates `addLiquidTag` & `addNunjucksTag` (no global varient)
- [ ] Do something about `markdownTemplateEngine`, `addMarkdownHighlighter`
- [ ] A whole bunch of stuff we haven't found yet

Not in this PR: configuration related methods: `setLiquidParameterParsing`, `setLiquidOptions`, `setNunjucksEnvironmentOptions`. We can't do this yet.

## Finishing touches

- [ ] Test code
- [ ] Fix tests

Linked Issues #3652, #3647. I've opened this pull request early in the process to make the intent clear

### DIFF
--- a/src/Engines/Nunjucks.js
+++ b/src/Engines/Nunjucks.js
@@ -6,6 +6,7 @@ import TemplateEngine from "./TemplateEngine.js";
 import EleventyBaseError from "../Errors/EleventyBaseError.js";
 import { augmentObject } from "./Util/ContextAugmenter.js";
 import { withResolvers } from "../Util/PromiseUtil.js";
+import isAsyncFunction from "../Util/IsAsyncFunction.js";
 
 const debug = debugUtil("Eleventy:Nunjucks");
 
@@ -21,6 +22,14 @@ export default class Nunjucks extends TemplateEngine {
 		this._usingPrecompiled = Object.keys(this.nunjucksPrecompiledTemplates).length > 0;
 
 		this.setLibrary(this.config.libraryOverrides.njk);
+		this.asyncFilters = eleventyConfig.getFilters({
+			lang: "njk",
+			async: true,
+		});
+		this.filters = eleventyConfig.getFilters({
+			lang: "njk",
+			async: false,
+		});
 
 		// v3.1.0-alpha.1 we’ve moved to use Nunjucks’ internal cache instead of Eleventy’s
 		// this.cacheable = false;
@@ -100,9 +109,8 @@ export default class Nunjucks extends TemplateEngine {
 		});
 
 		this.setEngineLib(this.njkEnv);
-
-		this.addFilters(this.config.nunjucksFilters);
-		this.addFilters(this.config.nunjucksAsyncFilters, true);
+		this.addFilters(this.filters);
+		this.addFilters(this.asyncFilters, true);
 
 		// TODO these all go to the same place (addTag), add warnings for overwrites
 		// TODO(zachleat): variableName should work with quotes or without quotes (same as {% set %})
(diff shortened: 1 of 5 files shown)


Ryuno-Ki:
How do you tell apart async from sync Nunjucks filter (in the summary)?

src/Engines/Nunjucks.js
```diff
@@ -121,7 +122,35 @@ export default class Nunjucks extends TemplateEngine {
...
+					let ret = filters[name].call(this, ...args);
+					if (ret instanceof Promise) {
+						throw new Error(
+							`Nunjucks *is* async-friendly with \`addFilter("${name}", async function() {})\` but you need to supply an \`async function\`. You returned a promise from \`addFilter("${name}", function() {})\`. Alternatively, use the \`addAsyncFilter("${name}")\` configuration API method.`,
+						);
+					}
+					return ret;
+				}
+			}
+			else if (isAsync) {
```
Ryuno-Ki, line 139: I could be wrong but it doesn't look like you need an `else if` because `isAsync` is Boolean already.

src/Engines/Nunjucks.js
```diff
@@ -121,7 +122,35 @@ export default class Nunjucks extends TemplateEngine {
...
 		for (let name in filters) {
-			this.njkEnv.addFilter(name, Nunjucks.wrapFilter(name, filters[name]), isAsync);
+			let callback = filters[name];
+			if (!isAsync) {
+				/** @this {any} */
+				callback = function (...args) {
+					// Note that `callback` is already a function as the `#add` method throws an error if not.
+					let ret = filters[name].call(this, ...args);
+					if (ret instanceof Promise) {
+						throw new Error(
```
Ryuno-Ki, line 132: Where is this supposed to be catched?

test/stubs/function-filter.11ty.cjs
```diff
@@ -1,4 +1,5 @@
 function myFunction({ name }) {
+  console.log("myFunction called",this);
```
Ryuno-Ki, line 2: Debug survivor.

Ryuno-Ki (commented):
I would suggest to start by moving the `*Dir` refactorings out into a separate PR.
They feel scoped enough that it should pass quickly without much ado.

Ryuno-Ki:
After reading the code I would have expected an

> Deprecates addAsyncNunjucksFilter in favour of `addFilter(…, …, { langs: ["njk"] }, true)`

above. Actually there are three possible values: `true | false | undefined` with the latter indicating a best-guess?

boehs:
Yes. Just a summary and actually incorrect. The marker overrides the detection

src/Engines/Nunjucks.js
boehs: It’s not. This is the same code addFilter added to nunjucks. It has been relocated to the nunjucks file for the purpose of decoupling

noelforte:
@boehs Adding a remark while its top of mind: While working on the Vento Plugin I've seen a lot of references and complex comparison chains to template language specific variables that could probably be consolidated down to a simpler API call (ie https://github.com/11ty/eleventy/pull/3785 that I just submitted). 

Not sure if you've identified some in your search already but figured it's worth pointing out so you can keep an eye out for them 😄

Zearin:
With 11ty v4 now in alpha, it looks like this is probably going to be “on hold” until v5 begins development.

That said, +1 for simplifying the API in this way! 👍

PR #3757, opened 2025-04-15
Initial `knip` improvements

zachleat:
https://knip.dev/

Recommended by @outslept: https://github.com/e18e/ecosystem-issues/issues/164#issuecomment-2807406743


outslept:
I’d recommend adding a [knip.json config file](https://knip.dev/reference/configuration) and specifying at least some ignore paths there. This will lower down the amount of unnecessary reports

zachleat:
Unfortunately I’ve decided against adding this one to the toolchain for now. I may revisit again in the future!

PR #3758, opened 2025-04-15
Migrate to flat config and drop .eslintrc support

outslept:
- migrated ESLint configuration to new flat config format
- removed usage of FlatCompat and .eslintrc* dependency
- switched stylistic rules to @stylistic/eslint-plugin-js


outslept:
Added an ignore field to silence unnecessary ESLint warnings in test files. This removes the bulk of the noise, but there are still a few lingering warnings left. Not sure yet how best to handle those, so I'll keep this PR as a draft for now

zachleat:
Ah, sorry — what is the benefit here? Seems like if I’m going to use ESLint it’s better to stick ESLint packages directly, no?

outslept:
> Ah, sorry — what is the benefit here? Seems like if I’m going to use ESLint it’s better to stick ESLint packages directly, no?

You're talking about the ignores field? I'm just worried about a big bunch of warnings after running lint command from root, so that's the workaround I've come up with.

If you don't need this I'm totally fine with removing it 😅

zachleat:
Ah, no sorry — I mean the use of `@stylistic` packages — I’m not familiar with this project

outslept:
> Ah, no sorry — I mean the use of `@stylistic` packages — I’m not familiar with this project

Ahh, with the ESLint v9 release, they just moved the stylistic rules out of the core and into separate stylistic plugin. So if you search for a rule like `space-unary-ops`, the documentation will show:

image

Source: https://eslint.org/docs/latest/rules/space-unary-ops

They deprecated these style rules from the core because they were becoming a massive maintenance burden, suggesting formatters or the @stylistic plugins as the new home for style rules.

zachleat:
Shipping with v4.0.0-alpha.1 thank you!

Issue #2844, opened 2023-03-05
Document how to access the Data Cascade in filters


mbforbes:
I was trying to use data from the Data Cascade in a filter. I asked about it in this discussion: https://github.com/11ty/eleventy/discussions/2841 and @pdehaan extremely helpfully solved it.

The relevant docs I could find are here:
https://www.11ty.dev/docs/filters/#scoped-data-in-filters

which say:
>A few Eleventy-specific data properties are available to filter callbacks.
> - `this.page` Added in v2.0.0
> - `this.eleventy` Added in v2.0.0
> ```js
> module.exports = function(eleventyConfig) {
>   // Make sure you’re not using an arrow function here: () => {}
>   eleventyConfig.addFilter("myFilter", function() {
>     // this.page
>     // this.eleventy
>   });
> };
> ```

However, I think this documentation could be improved:
1. There are more data properties available. For example, using Nunjucks, `this.ctx` is available. Critically, it contains the entire Data Cascade (I think). Perhaps we could list all the properties that are available?
2. There are properties available before v2.0.0. Because both the properties listed are tagged as v2.0.0, I thought those were the only ones. Since I was on v1.0.0, I didn't think the arrow function distinction would affect me, but it does.
3. The properties that are available depend on the filter. From @pdehaan, they're available at:
    - Nunjucks: `this.ctx`
    - Liquid: `this.context.environments`
    - Eleventy: (must pass in manually)

I think it's because I'm bad at JavaScript, but I find it very difficult to figure out what information I can access when I'm working inside filters or templating languages. Having documentation that lists out variables more exhaustively would help me out a lot. (I run into this genre of issue frequently working on my Eleventy + Nunjucks site.)

Many thanks for your consideration!

pdehaan:
> I think it's because I'm bad at JavaScript, but I find it very difficult to figure out what information I can access when I'm working inside filters or templating languages.

Here's a few tricks I use (apart from the obvious `console.log(this)`&hellip;

In most projects I end up using something like a custom `keys` filter which dumps all an object's available keys:
```js
eleventyConfig.addFilter("keys", obj => Object.keys(obj).sort());
```

Or one of my favorites:
```js
eleventyConfig.addFilter("inspect", obj => require("node:util").inspect(obj, {sorted: true}));
```

### USAGE
```njk
---
title: DEBUG
---

<h1>{{ title }}</h1>
<pre data-page-keys>{{ page | keys | inspect }}</pre>
[...]

PR #3785, opened 2025-04-28
Simplify retrieval of collections.all in `getLocaleCollectionItem`

noelforte:
This change motivated by: https://github.com/noelforte/eleventy-plugin-vento/issues/221

I couldn't figure out why the i18n plugin's presence was breaking the `getNextCollectionItem` and `getPreviousCollectionItem` filters since the Vento template engine plugin doesn't do anything fancy with the filters.

After a lot of trial and error, I discovered that this search for collections.all at the very end of the `GetLocaleCollectionItem` module checks a lot of things specific to Eleventy's own internal engines but of course custom engines don't add a `ctx` or `context` object, or even `this.collections`, since `this.page` and `this.eleventy` are the only context keys Eleventy supports, and `collections` is already accessible in page data anyway.

Rather than add `this.collections` to the Vento custom engine, could we simplify the `all` declaration to search the `data` key instead since the `data` key is already available on the render context anyway?

If not, I can look into adding `this.collections` to the custom engine for my plugin, but if this can be handled upstream that puts less pressure on custom engines to support adding it.

Thanks Zach!


noelforte:
Sorry to bug you Zach, any inclination on when this might get reviewed? Or is there something that I could do to help push things along?

noelforte:
@zachleat Just confirming, is this PR on hold for a specific reason? It's not been tagged as `on-hold`, and also I haven't heard anything about getting this reviewed apart from one reference in #2844. I know you're busy but I'd like to know whether I should be implementing a workaround on my own or whether the 11ty codebase can adapt these changes.

src/Filters/GetLocaleCollectionItem.js
```diff
@@ -36,11 +36,8 @@ function getLocaleCollectionItem(config, collection, pageOverride, langCode, ind
 	}
 
 	// find the modified locale `page` again in `collections.all`
-	let all =
-		this.collections?.all ||
-		this.ctx?.collections?.all ||
-		this.context?.environments?.collections?.all ||
-		[];
+	let all = this.collections?.all || this.data?.collections?.all || [];
```
noelforte, line 39: I'll need to confirm on my end, but I can't remember whether `this.data` is something provided by Eleventy, or something that is being injected by a custom engine in my use case. If `this.data` is Eleventy-supplied, then that's all that needs including in order to make custom engines pick up the locale `page` object.

The removal of lines 40-42 I did was mostly to reduce redundancy while (hopefully) not breaking backwards compatibility, but do let me know if I'm wrong!

noelforte: Can confirm, `this.data` is something provided by Eleventy since it appears even when a custom language isn't formatted.

cristovaov:
I came at this function from a different angle (pagination rather than a
custom engine) and ended up with a rig for it, so I ran this patch to see what
it does. Posting in case it helps move things along.

Applying the change as it stands, `all` comes back empty on every call that
reaches that line, on 3.1.6 and on 4.0.0-alpha.10. On my rig it goes from 6,
the whole collection, to 0. Every previous/next lookup that works today then
returns undefined. Nunjucks and Liquid both.

The reason looks like the two branches being dropped are the ones doing the
work. Measured on both versions:

| branch | Nunjucks | Liquid |
| --- | --- | --- |
| `this.collections?.all` | undefined | undefined |
| `this.data?.collections?.all` | undefined | undefined |
| `this.ctx?.collections?.all` | populated | undefined |
| `this.context?.environments?.collections?.all` | undefined | populated |

The filter context is `env, ctx, blocks, exported, page, eleventy` in Nunjucks
and `context, token, liquid, page, eleventy` in Liquid — no `data` in either.
I think `this.data` may be a Vento thing rather than a general one, which
would explain why this looks clean from that side.

Happy to share the rig if it is useful.

Issue #372, opened 2019-01-10
Nunjucks shortcodes without arguments throw an error (paired shortcodes work fine)


zachleat:
![image](https://user-images.githubusercontent.com/39355/50990605-189dd880-14d8-11e9-9008-c6091c5d474a.png)

PR #3789, opened 2025-05-01
fix: change missing arg from empty string to undefined

### LINKED ISSUE #3788: nunjucks shortcodes with no parameter passes empty string to function
### Operating system

Ubuntu 20

### Eleventy

3.0.0

### Describe the bug

If you add a shortcode with no parameter to a nunjucks template, it appears that the engine adds an empty string as a parameter. In other words, `{% test %}`is equivalent to `{% test "" %}`. This is not the behavior of liquid templates, where -- it appears -- `{% test %}`is equivalent to `{% test undefined %}`.

This might be related to #372. Hard to tell, because there's not much info in the issue, but I looked at the commit that resolved it (5cd19f6356c3601b4b6ddd50cfa026b569c3c83d), and it looks like the engine adds an empty string (cf. line 80+), apparently to resolve a nunjucks bug.

### Reproduction steps

1. Create a shortcode in an eleventy config file.
2. Call that shortcode from a nunjucks template with no parameter.
3. `console.log` the `typeof` the parameter.

### Expected behavior

In javascript, calling a function with no parameter passes undefined. So I think liquid's behavior is correct, and nunjucks' is not. However, maybe the nunjucks bug #372 prevents consistent behavior.

### Demo

https://github.com/btrem/11ty_3788

btrem:
May resolve #3788.

NB Needs testing.

### DIFF
--- a/src/Engines/Nunjucks.js
+++ b/src/Engines/Nunjucks.js
@@ -213,7 +213,7 @@ export default class Nunjucks extends TemplateEngine {
 				// though this issue is closed. Works fine for paired.
 				// https://github.com/mozilla/nunjucks/issues/158
 				if (args.children.length === 0) {
-					args.addChild(new nodes.Literal(0, 0, ""));
+					args.addChild(new nodes.Literal(0, 0, undefined));
 				}
 
 				parser.advanceAfterBlockEnd(tok.value);


btrem:
@zachleat I don't know how to test a branch of 11ty, so I can't check that what I've done doesn't break something, or reintroduce #372. (And even if I could test a branch, I'm not sure I'd know how to test the change I've made.)

zachleat:
You were close on this one! I think there is a better path forward at https://github.com/11ty/eleventy/pull/3822 though. Thanks for the effort!

Issue #3433, opened 2024-09-08
`addPreprocessor` doesn't work together with keyed `addExtension` 


mayank99:
### Operating system

macOS 14

### Eleventy

3.0.0-alpha.19

### Describe the bug

I'm trying to use `addPreprocessor` to transform the input file contents before handing it off to a template renderer.

This works for simple cases, but I've hit a wall when trying to use it with `key`ed `addExtension` (for [aliasing an existing template language](https://www.11ty.dev/docs/languages/custom/#aliasing-an-existing-template-language)). Specifically, I'm trying to use `key: "11ty.js"` so that I can access `this.defaultRenderer` inside the `compile` function. What happens is 11ty tries to process this template too early, before it has been transformed by the logic in `addPreprocessor`, which means it fails when it encounters unsupported syntax.

```
[11ty] Problem writing Eleventy templates:
[11ty] Unexpected token (3:4) (via SyntaxError)
```

### Reproduction steps

See [stackblitz](https://stackblitz.com/edit/github-qaud3e?file=eleventyJsxPlugin.js).

**Note**: I'm only using `"js"` to illustrate the problem more clearly. Using `"jsx"` or `"tsx"` also fails, but with a different "Unknown file extension" error (possibly related to https://github.com/11ty/eleventy/issues/3393#issuecomment-2276921293).

**Note 2**: A similar problem also happens in Deno (which natively supports JSX/TSX), except there is no error; the `content` in `addPreprocessor` is just empty. See [minimal repro](https://github.com/mayank99/11ty-jsx-deno/commit/421c20be51e078a0d8dd3363d3f1cd9f39910c92).

### Expected behavior

`addPreprocessor` should execute _before_ setting up a template alias (i.e. using `addExtension` with `key`). Both features should be usable together. 

### Reproduction URL

https://stackblitz.com/edit/github-qaud3e?file=eleventyJsxPlugin.js

zachleat:
Hmm. Agree that this is a bug. Currently we handle template types that read file content separate from dynamic `import` of JavaScript templates (through the `11ty.js` template type, aliased or not).

The most correct way to fix this is to restructure JavaScript templates to import during the file `read` step, which feels like a refactor that’s a little big for this late in the 3.0 release lifecycle.

This is the place where it forks:
* https://github.com/11ty/eleventy/blob/26c0fcfe7398c98651dfbff485fb7a14173e1aaa/src/TemplateContent.js#L269-L271
[...]

PR #3820, opened 2025-05-21
Fix JavaScript engine render so it always pulls fresh.

kobaj:
### Before we begin

I'm aware this doesn't currently pass tests, and is in fact at conflict with https://github.com/11ty/eleventy/issues/3433. I want to hopefully find some solution that makes both parties happy :).

### The problem I'm trying to solve

OS: Windows 11
Eleventy Version: 3.1.0
Node version: 20.10.0

Problem: Changes to dependencies are not reflected in the final compilation when running eleventy in `--serve` mode with jsx (or tsx) files.

### Steps to reproduce

I have a minimal example to reproduce the issue here of this available here: https://github.com/kobaj/minimal-eleventy-jsx-watch-example
1. Run `npm install`
2. Run `npm run serve`
3. Open browser to http://localhost:8080/
4. Edit the file `src/_includes/head.tsx`
    for example change "My test page, woo!" to be "Why isn't this working?"
5. Observe that the text in the header of your browser is still "My test page, woo!"

What I expect to happen at step 5 is for the header of my browser to be "Why isn't this working?". This pull request removes the caching happening inside eleventy and makes sure that changes to dependencies are reflected at render time. 

### Other things I've tried

I assumed this was a problem with dependency resolution. But the following didn't not fix my problem.

```javascript
// Placed inside of the tsx compile extension
this.addDependencies('./index.11ty.tsx', ['./_includes/head.tsx']);
```

### Where I'm at now

Again, I know that this pull request is not the best way to handle this, and breaks issue 3433. I would love your help in understanding what might be a better solution to my problem!

### DIFF
--- a/src/Engines/JavaScript.js
+++ b/src/Engines/JavaScript.js
@@ -1,9 +1,9 @@
-import { TemplatePath, isPlainObject } from "@11ty/eleventy-utils";
+import { isPlainObject } from "@11ty/eleventy-utils";
 
-import TemplateEngine from "./TemplateEngine.js";
 import EleventyBaseError from "../Errors/EleventyBaseError.js";
 import getJavaScriptData from "../Util/GetJavaScriptData.js";
 import { EleventyImport } from "../Util/Require.js";
+import TemplateEngine from "./TemplateEngine.js";
 import { augmentFunction, augmentObject } from "./Util/ContextAugmenter.js";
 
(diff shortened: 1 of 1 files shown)


zachleat:
Well, wait — this PR just removes caching altogether. Can you file an issue for the original issue?

kobaj:
Sure thing: https://github.com/11ty/eleventy/issues/3824

Issue #3938, opened 2025-09-11
Performance Optimization in Template Rendering


mohammed-hassan-angular-frontend-fcmb:
### Is your feature request related to a problem? Please describe.

Issue: There's a TODO comment about a hot inner loop in src/TemplateMap.js lines 328-336:

// TODO(slightlyoff): hot inner loop?
getMapEntryForInputPath(inputPath) {
    let absoluteInputPath = TemplatePath.absolutePath(inputPath);
    return this.map.find((entry) => {
        if (entry.inputPath === inputPath || entry.inputPath === absoluteInputPath) {
            return entry;
        }
    });
}

Performance Issues Identified

1. O(n) Linear Search: The method uses Array.find() which has O(n) time complexity
2. Called Frequently: This method is called in multiple hot paths:
- cache() method: Line 280 - called for every template in orderedPaths.map()
- populateCollectionsWithContent(): Line 482 - called for every collection item
- initDependencyMap(): Line 185 - called for every dependency entry
3. Various test scenarios and template processing workflows
Expensive Path Resolution: TemplatePath.absolutePath() is called on every lookup, even though the result could be cached
4. Scale Impact: With hundreds or thousands of templates, this becomes a significant bottleneck

### Describe the solution you'd like

Proposed Solution

Replace the linear search with a Map-based lookup system:

1. Add a lookup Map: Create this.inputPathMap alongside the existing this.map array
2. Maintain both structures: Keep the array for iteration, add the Map for O(1) lookups.
3. Handle both path formats: Store entries under both relative and absolute path keys.
4. Update on modifications: Ensure the lookup Map stays in sync when entries are added

### Describe alternatives you've considered

1. Constructor Changes
Add the lookup Map initialization:

constructor(eleventyConfig) {
    if (!eleventyConfig || eleventyConfig.constructor.name !== "TemplateConfig") {
        throw new Error("Missing or invalid `eleventyConfig` argument.");
    }
    this.eleventyConfig = eleventyConfig;
    this.map = [];
    
    // NEW: Add lookup Map for O(1) performance
    this.inputPathMap = new Map();
    
    this.collectionsData = null;
    this.cached = false;
    this.verboseOutput = true;
    this.collection = new TemplateCollection();
}

2. Enhanced add() Method
Modify the add method to populate both structures:

async add(template) {
    if (!template) {
        return;
    }

    let data = await template.getData();
    let entries = await template.getTemplateMapEntries(data);

    for (let map of entries) {
[...]

PR #3939, opened 2025-09-11
#3938 feat: optimize TemplateMap performance with O(1) lookup

mohammed-hassan-angular-frontend-fcmb:
- Replace O(n) Array.find() with O(1) Map.get() in getMapEntryForInputPath()
- Add inputPathMap for fast template lookups by input path
- Maintain backward compatibility with both relative and absolute paths
- Significant performance improvement for sites with many templates
- All existing tests pass

Fixes performance bottleneck in template rendering pipeline


zachleat:
Thanks!

PR #3956, opened 2025-10-24
Add optional `ELEVENTY_PERF=1` instrumentation to TemplateMap

j9t:
Adds timing measurements for `TemplateMap:cache`, `TemplateMap:initDependencyMap`, and `TemplateMap:setCollectionByTagName` when `ELEVENTY_PERF=1` is set.

Fell out of the bag after looking for performance improvements in my Eleventy projects—took the local Eleventy source as well and used this for tests. Proposing as it may provide useful beyond those local tests.

AI-generated context:

> This PR adds optional performance instrumentation to TemplateMap to help identify build bottlenecks. When enabled via `ELEVENTY_PERF=1` environment variable, it measures and reports timing for three key phases: `TemplateMap:cache` (overall caching phase), `TemplateMap:initDependencyMap` (dependency initialization), and `TemplateMap:setCollectionByTagName` (individual collection population).
>
> This instrumentation has zero performance overhead when disabled (the default), uses standard console.time/console.timeEnd APIs, and requires only ~15 lines of code. During testing on large sites (2,000+ files, 500+ collections), this revealed that template rendering—not collection population—accounts for the majority of build time, enabling targeted optimization efforts.

### DIFF
--- a/src/Data/TemplateData.js
+++ b/src/Data/TemplateData.js
@@ -305,7 +305,7 @@ class TemplateData {
 				);
 
 				let oldData = lodashGet(globalData, objectPathTarget);
-				data = TemplateData.mergeDeep(this.config.dataDeepMerge, oldData, data);
+				data = Merge(oldData, data);
 			}
 
 			dataFileConflicts[objectPathTargetString] = files[j];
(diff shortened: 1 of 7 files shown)


zachleat:
Going to close this one, sorry! Happy to chat about using/improving the existing benchmarking stuff we have internally though! https://github.com/11ty/eleventy/tree/f359aab2e76851f28d17e09b9f0debbf0fcee6c9/src/Benchmark

PR #4229, opened 2026-02-28
Perf: Memoize Render plugin's compile functions (#3915)

### LINKED ISSUE #3915: [Render plugin] Performance improvement opportunity with memoization of `renderFile` shortcode
### Is your feature request related to a problem? Please describe.

As mentioned in [another issue about the Bundle plugin bundling assets multiple times even with the same content](https://github.com/11ty/eleventy-plugin-bundle/issues/36), I recently [added memoization to the `compile` functions of my custom formats](https://github.com/nhoizey/eleventy-plugin-pack11ty/commit/9c55b72e1d204b26e0f2def559735e3199f62fbd) and improved my build time a lot.

Building my photography site on GitHub Actions used to take around 17 minutes, and it now takes less than 6 minutes (still a lot 😅, but much less).

Image

### Describe the solution you'd like

I wonder if memoization of [the `renderFile` shortcode](https://www.11ty.dev/docs/plugins/render/#render-file-shortcode) could be automated in the Render plugin, based on `inputPath` and/or a hash of `inputContent`.

### Describe alternatives you've considered

Every one can add memoization to his own code.

### Additional context

_No response_

agentic-sanyam:
Hey team,

This PR addresses #3915 by adding memoization to the Render plugin's `compile` and `compileFile` functions. 

The original issue highlighted that repeated calls to `renderFile` or `compile` with the same input content or file path could lead to redundant template compilations, especially on larger sites or during watch mode builds. As noted in the issue, implementing similar caching can lead to significant build time improvements, so I figured this would be a great addition to the plugin.

Here's a quick rundown of the changes:

*   **`compile` function:** Now uses a `WeakMap`-based `compileCache` to store and retrieve already compiled template functions. The cache key is generated from the template content and language, meaning if you `compile` the same string multiple times, it only gets processed once.
*   **`compileFile` function:** Similarly, `compileFile` now leverages a `WeakMap`-based `compileFileCache`. This cache stores the compiled function along with the file's content at the time of compilation. Before returning a cached function, it performs a quick check against the current file content to ensure the cache is invalidated if the file has changed (crucial for watch mode or other file modifications).
*   Both caches are scoped using the `TemplateConfig` object as a `WeakMap` key. This helps ensure proper isolation between different Eleventy instances and allows for efficient garbage collection when a config is no longer needed.

These changes should lead to a nice performance boost for sites that heavily use `renderFile` or `compile`, as it avoids re-compiling the same templates over and over. 

I've also added a few new tests to `TemplateRenderPluginTest.js` to verify that the caching works as expected, including correctly invalidating the cache when file contents change.

Let me know what you think!

Fixes #3915

### DIFF
--- a/src/Plugins/RenderPlugin.js
+++ b/src/Plugins/RenderPlugin.js
@@ -15,6 +15,9 @@ import Liquid from "../Adapters/Engines/Liquid.js";
 
 class EleventyNunjucksError extends EleventyBaseError {}
 
+const compileCache = new WeakMap();
+const compileFileCache = new WeakMap();
+
 /** @this {object} */
 async function compile(content, templateLang, options = {}) {
 	let { templateConfig, extensionMap } = options;
@@ -29,6 +32,15 @@ async function compile(content, templateLang, options = {}) {
 	// Breaking change in 2.0+, previous default was `html` and now we default to the page template syntax
 	if (!templateLang) {
 		templateLang = this.page.templateSyntax;
+	}
+
+	if (!compileCache.has(templateConfig)) {
+		compileCache.set(templateConfig, new Map());
+	}
+	let projectCache = compileCache.get(templateConfig);
+	let cacheKey = `${content}###${templateLang || ""}`;
+	if (projectCache.has(cacheKey)) {
+		return projectCache.get(cacheKey);
 	}
 
 	if (!extensionMap) {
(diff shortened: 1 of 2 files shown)


Ryuno-Ki:
Note: Account was created two days ago. The activity looks a little bit sus too me.

zachleat:
Stale.

Ryuno-Ki:
What kind of PRs do you mark with `stale`? Is there some resolution (e.g. reopen once rebased?) Or do you not plan to accept the PR even if were unstale?

Issue #4194, opened 2025-12-12
Add first-party support for TypeScript (type stripped) Configuration Files, Data Files, and Template Types


zachleat:
From #3616 and #4007

~~In preparation for Node.js stable to ship TypeScript stripping as an unflagged feature~~

Node.js has shipped type stripping in Node 22.6: https://nodejs.org/api/typescript.html

We already provide support for TypeScript in Eleventy v3.x via Node.js Type Stripping, Deno, and tsx, documented here: https://www.11ty.dev/docs/languages/typescript/

## Default Configuration File Names

- `eleventy.config.ts`
- `eleventy.config.mts`
- `eleventy.config.cts`
- _Not_ `.tsx`

## Default Data File Names

- `*.11tydata.ts`
- `*.11tydata.mts`
- `*.11tydata.cts`
- _Not_ `.tsx`

## Default TypeScript Template File Names

- `*.11ty.ts`
- `*.11ty.mts`
- `*.11ty.cts`
- _Not_ `.tsx`

zachleat:
Weeeee! Working example in https://github.com/11ty/eleventy/tree/4194

AleksandrHovhannisyan:
Very much looking forward to this! I got TypeScript working on Node 25 + eleventy 3.1.2 out of the box on a branch for my site, but data files currently seem to get ignored. I assume that's expected and that's what this issue is for?

Edit: looking at the commit above, I think it's expected that it doesn't currently work (but will). Thanks for working on this!

SignpostMarv:
> Very much looking forward to this! I got TypeScript working on Node 25 + eleventy 3.1.2 out of the box on a branch for my site, but data files currently seem to get ignored. I assume that's expected and that's what this issue is for?
> 
> Edit: looking at the commit above, I think it's expected that it doesn't currently work (but will). Thanks for working on this!

Have got typescript data working on node 24 thusly:

```ts
	config.addDataExtension('ts', {
		read: false,
		parser: async (filepath: string): Promise<unknown> => {
			const real = await realpath(filepath);

			// sanity check that data is where you expect it to be
			if (!(real.startsWith(`${import.meta.dirname}/11ty/data/`))) {
				throw new Error('Unsupported data file!');
			}

			return (await import(real) as {default: unknown}).default;
		},
	});
```

AleksandrHovhannisyan:
@SignpostMarv Just tried that and it works, thank you!

[...]

PR #4247, opened 2026-03-28
Fix TypeScript data files not being processed

cypressSap:
As noted in https://github.com/11ty/eleventy/issues/4194#issuecomment-4106151288, TypeScript data files don't yet work in v4.0.0-alpha.7. This is a bit quick and dirty but seems to get them going, both global `*.ts` (etc.) and directory `*.11tydata.ts` (etc.) files :)


src/Data/TemplateData.js
```diff
@@ -508,7 +508,14 @@ class TemplateData {
...
 		let extension = TemplatePath.getExtension(path);
 
-		if (extension === "js" || extension === "cjs" || extension === "mjs") {
+		if (
+			extension === "js" ||
+			extension === "cjs" ||
+			extension === "mjs" ||
+			extension === "ts" ||
+			extension === "cts" ||
+			extension === "mts"
```
Ryuno-Ki: At this point, a slightly different approach could be more readable:

```js
["js", "cjs", "mjs", "ts", "cts", "mts"].some(extension)
```

src/Data/TemplateData.js
```diff
@@ -576,6 +588,9 @@ class TemplateData {
 				paths.push(base + suffix + ".js");
 				paths.push(base + suffix + ".cjs");
 				paths.push(base + suffix + ".mjs");
+				paths.push(base + suffix + ".ts");
+				paths.push(base + suffix + ".cts");
+				paths.push(base + suffix + ".mts");
```
Ryuno-Ki: Again, having an array of the suffixes and use a `.forEach()` to handle the `push` could be easier to read.

Ryuno-Ki (commented):
I don't see new files being added. But then I'm not deep enough into the codebase to tell whether that's expected.

src/Data/TemplateData.js
cypressSap: Thanks, cleaned these both up :+1:

cypressSap:
> I don't see new files being added. But then I'm not deep enough into the codebase to tell whether that's expected.

Sorry, I'm not sure what new files this is referring to?

test/TemplateDataTest.js
```diff
@@ -380,12 +380,18 @@ test("getLocalDataPaths", async (t) => {
   t.deepEqual(paths, [
     "./test/stubs/stubs.json",
     "./test/stubs/stubs.11tydata.json",
+    "./test/stubs/stubs.11tydata.mts",
```
Ryuno-Ki, line 383: This sounds like a new file ought to be added.

cypressSap: Oh, gotcha. Most of the files already listed don't exist either, though.

Ryuno-Ki: Therefore:

> But then I'm not deep enough into the codebase to tell whether that's expected.

zachleat (approved):
Appreciate this contribution!

zachleat:
This will ship with v4.0.0-alpha.8.

zachleat:
Notably, I did add an `ELEVENTY_SKIP_TYPESCRIPT` opt out environment variable to opt out of TypeScript behavior!

Ryuno-Ki:
> `ELEVENTY_SKIP_TYPESCRIPT`

I don't see it documented in the website repo (or am I missing something?)

zachleat:
You’re right, I’ll add a TODO label here

PR #4290, opened 2026-06-04
Fix mixed-case custom extension matching

puneetdixit200:
## Summary
- normalize custom `addExtension()` extension names and keys before storing them
- make extension-map path checks compare extensions case-insensitively
- add regression coverage for a mixed-case `vZome` custom extension and duplicate mixed-case registrations

Fixes #4224.

## Tests
- `npx ava --verbose test/EleventyExtensionMapTest.js`
- `npx ava --verbose test/EleventyTest.js --match='*issue #4224*'`
- `npx prettier --check src/UserConfig.js src/EleventyExtensionMap.js test/EleventyExtensionMapTest.js test/EleventyTest.js`
- `npx eslint src/UserConfig.js src/EleventyExtensionMap.js test/EleventyExtensionMapTest.js test/EleventyTest.js` (passes with the existing `src/UserConfig.js` unused-argument warning)
- `npm run check` (passes with existing warnings)
- `npm run test:server` (1417 passed, 20 skipped)
- `npm run test:client` fails locally because Playwright browser executables are missing from `~/Library/Caches/ms-playwright`; Vitest reports to run `npx playwright install`. The commit/push hooks hit the same local browser-cache blocker after the server suite passed.

### DIFF
--- a/src/EleventyExtensionMap.js
+++ b/src/EleventyExtensionMap.js
@@ -76,6 +76,8 @@ class EleventyExtensionMap {
 	// on paths found from the file system glob search.
 	// TODO: Method name might just need to be renamed to something more accurate.
 	isFullTemplateFilePath(path) {
+		path = (path || "").toLowerCase();
+
 		for (let extension of this.validTemplateLanguageKeys) {
 			if (path.endsWith(`.${extension}`)) {
 				return true;
@@ -97,6 +99,8 @@ class EleventyExtensionMap {
 	}
 
 	getValidExtensionsForPath(path) {
+		path = (path || "").toLowerCase();
+
 		let extensions = new Set();
 		for (let extension in this.extensionToKeyMap) {
 			if (path.endsWith(`.${extension}`)) {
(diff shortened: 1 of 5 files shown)


src/EleventyExtensionMap.js
```diff
@@ -76,6 +76,8 @@ class EleventyExtensionMap {
 	// on paths found from the file system glob search.
 	// TODO: Method name might just need to be renamed to something more accurate.
 	isFullTemplateFilePath(path) {
+		path = (path || "").toLowerCase();
```
Ryuno-Ki, line 79: Generally speaking: it's an anti-pattern to modify function arguments.
Can you use a new variable like the normalizedPath below?

test/stubs-4224/index.vZome
```diff
@@ -0,0 +1 @@
+mixed-case custom extension
```
Ryuno-Ki, line 1: What's the purpose of this file?

test/EleventyTest.js
```diff
@@ -1622,6 +1622,26 @@ test("permalink on custom template lang, issue #3619", async (t) => {
 }`);
 });
 
+test("Mixed-case custom template extensions are discovered, issue #4224", async (t) => {
```
Ryuno-Ki, line 1625: Ah, okay, so the content is less of importance. Gotcha.

